PR #9, opened 2016-08-15
Properly bubble up exceptions that are thrown by the parsing code

anthonycr:
#### Ticket
[VA-1894](https://vimean.atlassian.net/browse/VA-1894)

#### Ticket Summary
Due to some buggy json we are pulling down from the server, we are getting a crash in the android app. We want to pinpoint the exact part of the object that isn't able to be parsed.

#### Implementation Summary
We should be bubbling up exceptions in our parsing code so that the consumer can track down the precise location of the parsing failure.

#### How to Test
See the sample app, modify one of the fields that isn't a Date in a model class to a Date and of course it will fail with a parse exception. Now we will bubble up the origin instead of just catching an error

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,7 @@ MyParsingClass {
 ## Future Enhancements
 
 - Generate code so that member variables only need to be package local
+- Add an option to absorb parsing errors rather than re-throwing them
 
 ## License
 `stag-java` is available under the MIT license. See the [LICENSE](LICENSE) file for more information.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -127,6 +127,7 @@ MyParsingClass {
 ## Future Enhancements
 
 - Generate code so that member variables only need to be package local
+- Add an option to absorb parsing errors rather than re-throwing them
```
anthonycr: I think this would be a good option. We can discuss it, but it could be done on a class by class basis or globally. Basically if we hit an IOException we would check if the user wants to be notified by rethrowing it, otherwise we would swallow it and proceed with parsing.

kvenn:
Looks good, we should get something in place for tracking library related tickets. I'll bring it up in Retro.

Merge when ready! 💯

PR #10, opened 2016-08-16
use google auto service

penkzhou:
#### Ticket Summary
use google auto service 

#### Implementation Summary
add google auto service to generate META-INF metadata for Java
annotation processors automatically

#### How to Test
sync the gradle and stag-java-compiler generate the same source java file.


anthonycr:
Great addition, thanks!

PR #14, opened 2016-10-19
Improvement to allow operation across modules

mcumings:
Curious whether you folks would be open to this contribution which refactors
Stag to allow for annotation processing to span multiple modules.  The application
I am working on has components located within a library module as well as components
located int he main application module.  As it currently stands, the generated Stag.java
does not expose the library module's generated TypeAdapter instances.  This PR is
intended to address this while retaining Stag.Factory API compatibility.

=== Change description is as follows:

Previously, the generated Stag class would be directly encoded with the
types it knew about at compilation-time seeded with information about the
classes which were annotated (in the current pass only).  When code was
generated for entities in multiple modules this resulted in the Stag class
only referencing classes generated in the outermost module.  In order to
retain knowledge of generated classes across modules and annotation
processing passes, the KnownTypeAdapterFactoriesUtils was added which
maintains a list of all known TypeAdapterFactories in a StagTypeAdapterFactory.list
resource, ServiceLoader style.

Extracted and distributed the parsing logic into each type adapter
implementation, making them fully independent and eliminating a Gson
TypeAdapter cache lookup upon each non-native field access.  In doing
so also fixed a bug where only writer.beginObject was called for null
object values.

Relocates the generated Adapters classes into the context of their originating
package to eliminate naming conflicts.

(Re) adds support for nested model classes.

Note: It would be trivial to adapt this to a pure ServiceLoader approach where
META-INF/services/com.google.gson.TypeAdapterFactory is used in place of
StagTypeAdapterFactory.list but this approach does not currently work on
Android due to toolchain bugs.


anthonycr:
Checking this out, from your description this seems like it would be a good addition. I'll dig into it

anthonycr:
A side effect of this PR is that one of the requirements I've been wanting to get rid of (public visibility on member variables) has been removed. Since it's no longer necessary, could you make a small change?

In the `StagProcessor` class, the following check

```java
} else if (!modifiers.contains(Modifier.PUBLIC)) {
    throw new RuntimeException("Unable to access field \"" +
        variableElement.getSimpleName().toString() + "\" in class " +
        variableElement.getEnclosingElement().asType() +
        ", field must public.");
}
```

can be changed to

```java
} else if (modifiers.contains(Modifier.PROTECTED) || modifiers.contains(Modifier.PRIVATE)) {
    throw new RuntimeException("Unable to access field \"" +
        variableElement.getSimpleName().toString() + "\" in class " +
        variableElement.getEnclosingElement().asType() +
        ", field must have package or public visibility.");
}
```

This way, member variables won't need to be exposed as public any longer 😎

mcumings:
I can absolutely make the requested change.  Glad to accidentally address your desire! ;)

To clarify though, the fields must now just not be PRIVATE, correct?  i.e.,  PUBLIC, PROTECTED, and DEFAULT will all work.

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/StagGenerator.java
```diff
@@ -102,111 +95,50 @@ public void generateTypeAdapters() throws IOException {
...
-                         "\tlist.add(typeAdapter.read(in));\n" +
-                         "}\n" +
-                         '\n' +
-                         "return list;\n")
-                .build();
-    }
-
-    @NotNull
-    private static TypeSpec getAdapterFactorySpec(Set<Element> types) {
+    private TypeSpec getAdapterFactorySpec() {
```
mcumings: My personal preference is to take SRP to its limit and keep every individual class fully
separated.  The TypeAdapter and TypeAdapterFactory have different purposes that can
be composed in the consuming application in various ways.  If the two are collapsed
together then this flexibility starts to diminish and the consumer is boxed into a corner.

The second option would make the class instances stateful which would be more difficult
to understand as a casual observer.  Also, care would need to be taken with respect to
the memory model in this scenario as - I believe - Gson can initialize type adapters in
separate threads.

Not my call though. 😄

stag-library-compiler/src/main/java/com/vimeo/stag/processor/StagProcessor.java
```diff
@@ -120,15 +128,32 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
...
             for (Entry<Element, List<VariableElement>> entry : variableMap.entrySet()) {
                 SupportedTypesModel.getInstance()
                         .addSupportedType(new AnnotatedClass(entry.getKey(), entry.getValue()));
             }
-            ParseGenerator parseGenerator = new ParseGenerator(mSupportedTypes, processingEnv.getFiler());
-            parseGenerator.generateParsingCode();
-            StagGenerator adapterGenerator = new StagGenerator(processingEnv.getFiler());
+            mSupportedTypes.addAll(KnownTypeAdapterFactoriesUtils.loadKnownTypes(processingEnv));
+
+            StagGenerator adapterGenerator = new StagGenerator(filer, mSupportedTypes);
```
mcumings: That's correct.

In my scenario I have an Android application module that includes an Android library, both
of which define entities.  I was a little worried about having the Stag class generated twice
but the Android toolchain seems to do the right thing and use the application module
version.

Ideally I think we'd leverage ServiceLoader from within a non-generated Stag factory class
in the stag-library module.  But...  Android is eating META-INF entries created in the
application module and putting the Stag class under stag-library would require creating
a dependency reference to Gson.  The latter can be handled carefully so as to not force
library versions on folks, etc, but it is still a detractor.

mcumings:
Added two more commits integrating the PR feedback thus far.

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/StagGenerator.java
anthonycr: excellent point. Flexibility is important for a library. I'm fine with this then

anthonycr:
Looks awesome, going to merge it!

mcumings:
Awesome!  Thanks!

mcumings:
Hey @anthonycr - would it be possible to spin up a release build now that this has merged to master?  It would save me from having to plumb up a non-official binary inside my company.

anthonycr:
@mcumings sorry for the delay, but I just released v1.1.0 👍

PR #18, opened 2016-11-07
List adapter field fix

kevinzetterstrom:
#### Ticket Summary
There is an issue with List types in that the actual parameterized type is being added as an adapter field, even if that type exists. For most types this is already prevented by using a HashSet, however, lists can get around this.

#### Implementation Summary
I created a second HashSet, and iterate over the first to create the second, then use the second to add the adapter fields. I also added a try/catch around ArrayList inner read calls to be in line with other object read calls.

Bumping version to 1.1.1. I will update the readme once it this is approved before deployment.

#### How to Test
I changed the Video model to have a bunch of duplicate types and ArrayLists. Verify the Video TypeAdapter is correctly generated.

### DIFF
--- a/sample/src/main/java/com/vimeo/sample/model/Video.java
+++ b/sample/src/main/java/com/vimeo/sample/model/Video.java
@@ -25,12 +25,25 @@ package com.vimeo.sample.model;
 
 import com.vimeo.stag.GsonAdapterKey;
 
+import java.util.ArrayList;
 import java.util.Date;
 
 public class Video {
 
     @GsonAdapterKey("user")
     public User mUser;
+
+    @GsonAdapterKey("secondUser")
+    public User mSecondUser;
+
+    @GsonAdapterKey
+    ArrayList<User> mUsersList;
+
+    @GsonAdapterKey
+    ArrayList<User> mSecondUsersList;
+
+    @GsonAdapterKey
+    public ArrayList<Stats> mStatsArrayList;
 
     @GsonAdapterKey("link")
     public String mLink;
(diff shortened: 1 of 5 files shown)


sample/src/main/java/com/vimeo/sample/model/Video.java
```diff
@@ -25,13 +25,26 @@
...
+
+    @GsonAdapterKey
+    ArrayList<User> mUsersList;
+
+    @GsonAdapterKey
+    ArrayList<User> mSecondUsersList;
+
+    @GsonAdapterKey
+    public ArrayList<Stats> mStatsArrayList;
+
```
anthonycr: Minor: could we move this to a separate class since these added fields aren't actually part of the Video object returned from the server? Maybe just call it `IdenticalFieldTypes` or something

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
```diff
@@ -288,13 +294,20 @@ private static String getReadTokenType(@NotNull TypeMirror type) {
...
                    prefix + "object." + variableName + " = new java.util.ArrayList<>();\n" +
                    prefix + "while (reader.hasNext()) {\n" +
-                   prefix + "\tobject." + variableName + ".add(" + innerRead + ");\n" +
+
+                   prefix + "\ttry {\n" +
+                   prefix + "\t\tobject." + variableName + ".add(" + innerRead + ");\n" +
+                   prefix + "\t} catch(Exception exception) {\n" +
+                   prefix + "\t\tthrow new IOException(\"Error parsing " +
+                   outerClassType + "." + variableName + " JSON!\", exception)\n;" +
+                   prefix + "\t}" +
```
anthonycr: 👍

anthonycr:
m e r g e   i t

PR #20, opened 2016-11-15
Support for TypeTokens for complex Data Structures and Performance Improvements

anirudhramanan:
Code Changes Description :

**1. Support for TypeToken which will allow more complex types to be supported by Stag** : We added the support for typetokens which will now allow complex types such as HashMap, ArrayList<ArrayList> etc, to be supported. These typetokens will be created as a static final member variable in StagTypeTokenConstants.java file, while keeping a reference to the generated typetokens so that we do not create the typetoken of same type twice.

Performance Improvements : 

Previously, in the TypeAdapter constructor, TypeToken.get(type) was being used. This internally creates a new typetoken for the type. Since creating new typetokens are really cpu intensive, this could lead to performance degradation. This was solved by keeping the reference to the typetokens that were generated in the StagTypeTokenConstants.java file and reusing them.

Testing : Use HashMaps or any other complex data type as a field in the class to be generated.

**2. Support for float data type**

Testing : Create a float variable in a class.

**3. Change in name convention for generating typeadapter files** : In case a class of same name is present in same package as a static inner class and an outer class as well, this resulted in generation of typeadapter files with same name. 

Repro : Added a class with same name as an outer class and as an inner static class of some other parent class. This will throw a build time error since there are two typeadapters with same name.

Testing : Same as repro steps.

**4. Removed try/catch block for nested cases** : Since try/catch exception blocks decreases performance, we removed these blocks especially for nested cases. Given that the core philosophy of Stag is based on performance all the current try/catch blocks give is more debugging ability. 

Testing : More of a performance gain.

**5. Fixed IllegalStateException while writing** : We're able to repro this while writing nested contents using stag. This was fixed as follows

Previously, while writing, if the object to be written was null, the writer.beginObject() and endObject() were not called, instead the method was returned. With the new changes, writer.beginObject() is called before the null check for object. In case the object is null, the writer.endObject() is called and method is returned.

### DIFF
--- a/sample/src/main/java/com/vimeo/sample/model/NestedModel.java
+++ b/sample/src/main/java/com/vimeo/sample/model/NestedModel.java
@@ -1,0 +1,14 @@
+package com.vimeo.sample.model;
+
+import com.google.gson.annotations.SerializedName;
+
+/*
+    This is to simulate a scenario where a class with the same name is present as
+    a static inner class for another class in the same package
+ */
+
+public class NestedModel {
+
+    @SerializedName("test1")
+    public String test1;
+}
--- a/sample/src/main/java/com/vimeo/sample/model/NestedModel.java
+++ b/sample/src/main/java/com/vimeo/sample/model/NestedModel.java
@@ -1,6 +1,7 @@
 package com.vimeo.sample.model;
 
 import com.google.gson.annotations.SerializedName;
+import com.vimeo.stag.GsonAdapterKey;
 
 /*
     This is to simulate a scenario where a class with the same name is present as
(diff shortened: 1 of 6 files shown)


sample/src/main/java/com/vimeo/sample/model/NestedModel.java
```diff
@@ -0,0 +1,14 @@
...
+import com.google.gson.annotations.SerializedName;
+
+/*
+    This is to simulate a scenario where a class with the same name is present as
+    a static inner class for another class in the same package
+ */
+
+public class NestedModel {
+
+    @SerializedName("test1")
```
anirudhramanan: pushed the review changes in

PR #24, opened 2016-11-18
Multiple Support to Stag 

anirudhramanan:
Code Changes :

- **Package Name Support** :  Now you can pass the path to generate the stag files as an argument to the apt compiler. This will help in scenarios where multiple jars that are being used has Stag generated code.

- **Removal of Separate TypeAdapterFactories for each model** : Generation of separate TypeAdapterFactories resulted in increase in number of methods. In our scenario, where we had 300+ models, it resulted in 300+ TypeAdapterFactories which bumped the method counts. Now since all the factories have been moved inside stag.factory we were able to save around **300 kb.**

   > The change of generating separate TypeAdapterFactories was made to support stag across modules. Now since we have package name support to generate files in diff package, this will not break the support for the same.

  > Since a single method declaration results in generation of 2-3 methods internally in java, this change will help save lot of methods. Reference : www.realm.io/news/360andev-jake-wharton-java-hidden-costs-android/

- **Support for Generic/Parameterized Classes** : Stag now supports the parsing of generic / parameterized classes. For the generic types, ParameterizedTypeUtil class will be generated to get its corresponding typetoken for parsing. For rest of the concrete types, parsing will be done as usual.

- **Support for Android Support Annotations (like NonNull)** : If any field is marked as NonNull using android annotations, it will throw an IOException if it is null. The exception is internally caught while parsing, hence will not crash the app.


mcumings:
This restores the multi-module build capabilities after being broken in release 1.1.2 due to the type tokens being stored in a static location across all models.  I separately developed this change with the intent to contribute and then only later saw this PR.  If I could request that this get merged and released soon it would help me out greatly...

As an aside, the type token code will still fail in multi-round or incremental compilation scenarios as the SupportedTypesModel is populated only with type knowledge gathered from a single processing round.  That was one of the primary reasons why the out-of-band KnownTypeAdapterFactoriesUtils was created in the original multi-module commit for 1.1.0.

yasirmhd:
@mcumings Thanks for reporting the issues of incremental builds with KnownTypeAdapterFactoriesUtils . We will work on fixing that in a subsequent PR

anirudhramanan:
@anthonycr : These changes will add much more capabilities to stag as a library.

mcumings:
We should probably make the KnownTypeAdapterFactoryUtils utilize the specified package name as a dynamic component of its KNOWN_FACTORIES_RESOURCE instead of a using a static value.  Since multi-module support is now achieved through specification of a unique package name, we don't want knowledge of types to span modules through a single resource.  If we don't then library model definitions will be re-included in application-level definitions of Stag but without the knowledge of the relevant type tokens.

anirudhramanan:
@mcumings I assume you are talking about generating different KNOWN_FACTORIES_RESOURCE for different modules. Since KnownTypeAdapterFactoryUtils takes in the package name for the file to be written, each module will have its KNOWN_FACTORIES_RESOURCE consisting of its own types.

mcumings:
Ah, d'oh!  I saw the commit restoring it and made the bad assumption that it was being restored to its original form.  My bad, sorry.  It should work just fine for what I was after.

Thanks!

anthonycr:
sorry I'm just getting around to this, going to dig in and review now

stag-library-compiler/src/main/java/com/vimeo/stag/processor/StagProcessor.java
```diff
@@ -104,27 +114,27 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
...
                 }
 
                 Element enclosingClassElement = variableElement.getEnclosingElement();
                 TypeMirror enclosingClass = enclosingClassElement.asType();
 
-                if (!TypeUtils.isParameterizedType(enclosingClass) ||
-                    TypeUtils.isConcreteType(enclosingClass)) {
-                    mSupportedTypes.add(enclosingClass.toString());
+                if (TypeUtils.isParameterizedType(enclosingClass) ||
+                        TypeUtils.isConcreteType(enclosingClass)) {
```
anthonycr: I don't think this check is necessary anymore, since the `isParameterizedType` call checks whether the type is parameterized, and the `isConcreteType` checks whether either the class has no type parameters, or the type parameters that are there are resolved to real types.

anthonycr: There is also new bug that's introduced with the support for instantiating these non-concrete parameterized types. In our network layer, we utilize abstract parameterized classes that are never instantiated, but only extended by other model objects. The current code tries to mistakenly instantiate these abstract model objects.

I should have made the `SuperAbstractDataList` class in the sample app `abstract` so that this situation would have come up in testing, but I made a mistake and forgot about the possibility.

If we can prevent a type adapter from being created for abstract classes (by checking the modifier before calling `mSupportedTypes.add` and `addToListMap` in this processor class) then the bug will be fixed.

So two things:
1. Can you make the `SuperAbstractDataList` class abstract
2. Can you add in checks to make sure type adapters are not generated for abstract classes

anthonycr: basically, this bug already existed (e.g. if you had a non parameterized abstract model class), but it's just becoming apparent now

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
```diff
@@ -173,14 +209,40 @@ private static MethodSpec getWriteMethodSpec(@NotNull TypeName typeName,
...
+    @NotNull
+    private static String getTypeTokenCode(@NotNull TypeMirror fieldType, @NotNull Map<TypeVariable, String> typeVarsMap, @NotNull TypeTokenConstantsGenerator typeTokenConstantsGenerator) {
+        String result = null;
+        if (!TypeUtils.isConcreteType(fieldType)) {
+            if (fieldType.getKind() == TypeKind.TYPEVAR) {
+                result = "(" + ClassName.get(TypeToken.class) + "<" + fieldType + ">)" + ClassName.get(TypeToken.class) + ".get(" + typeVarsMap.get(fieldType) + ")";
+            } else if (fieldType instanceof DeclaredType) {
+                DeclaredType decaredFieldType = (DeclaredType) fieldType;
+                List<? extends TypeMirror> typeMirrors = ((DeclaredType) fieldType).getTypeArguments();
+                result = "com.vimeo.sample.stag.generated." + ParameterizedTypeGenerator.CLASS_NAME + ".getTypeToken(" + decaredFieldType.asElement().toString() + ".class";
```
anirudhramanan: Yes, the package name should have been dynamic. My Bad

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/ParameterizedTypeGenerator.java
```diff
@@ -0,0 +1,185 @@
...
+
+        adapterBuilder.addField(fieldSpec);
+        adapterBuilder.addMethod(constructorBuilder.build());
+        adapterBuilder.addMethod(ownerTypeMethod.build());
+        adapterBuilder.addMethod(rawTypeMethod.build());
+        adapterBuilder.addMethod(typeArgumentMethod.build());
+        return adapterBuilder.build();
+    }
+
+    public static void generateParameterizedUtilClass(Filer filer, String packageName) throws IOException {
```
yasirmhd: Agree with that. We will move this to stag-library. Does that work?

stag-library-compiler/src/main/java/com/vimeo/stag/processor/StagProcessor.java
yasirmhd: We will push some more changes with this PR later in the day today. We will fix these. We also did some profiling and saw a bunch of improvements we can do by not creating TypeTokens constants if not absolutely necessary.

anthonycr: great, thank you!

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
anirudhramanan: Since the ParameterizedTypeUtils has been moved to the stag-library, this is not required anymore.

yasirmhd:
@anthonycr  : The PR is good to go now. Kindly review.

anthonycr:
I'm seeing that if you make the `SuperAbstractDataList` class `abstract`, the fields are no longer resolved and included in the `VideoList` type adapter. It doesn't crash anymore, but it doesn't recurse the hierarchy like before and find the fields.

yasirmhd:
@anthonycr if you can please share the code where it crashes or the call stack we will work on fixing it.

yasirmhd:
@anthonycr  I have fixed it now. Can you please check it out?

yasirmhd:
@anthonycr : Kindly review 	d967d4f as well. This will be the last of the optimisation changes :-)

anthonycr:
This looks great, tested everything and it's 👌 

merging

mcumings:
@anthonycr - Can you please spin up an official release containing this improvement?

anthonycr:
@mcumings I'll try to get it out by the end of the week, unfortunately I was a bit too busy the past several days to get it done

mcumings:
Okay, sounds great.  Appreciate it!

PR #26, opened 2016-12-08
Parsing Maps and Enums the Stag way + bump Gson 2.8.0 API

anirudhramanan:
Code Changes : 

1. Support of Parsing Maps and Enums : Added support for parsing maps and enums similar to other data types such as list etc. Since the adapters used by Gson uses reflection, this was slowing down the parsing. 

2. With Gson 2.8.0, the parameterized util function has been made public, hence we can remove the util method that was added to the stag-library.

screen shot 2016-12-08 at 4 21 04 pm


yasirmhd:
@anthonycr : This is a much better implementation of Maps as the adapters by Gson use reflection (https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/internal/bind/MapTypeAdapterFactory.java)

sample/src/main/java/com/vimeo/sample/model/ClassWithMapTypes.java
```diff
@@ -0,0 +1,26 @@
...
+    public LinkedHashMap<String, Integer> stringIntegerMap;
+
+    @GsonAdapterKey
+    public Map<Integer, String> integerStringMap;
+
+    @GsonAdapterKey
+    public Map<Video, String> complexTypeMap;
+
+    @GsonAdapterKey
+    public Map<String, List<Video>> mapOfLists;
```
yasirmhd: Yes, Maps of Maps is also something which is on our radar. Want to do it efficiently without increasing the number of methods and classes. Still thinking of an optimal solution for this.

PR #30, opened 2016-12-20
Basic for loops in write adapter code

anthonycr:
#### Summary
I noticed we could make a small memory optimization in the list generation code in the write method of the type adapter.

#### Implementation Summary
The optimization was to switch from an enhanced for loop to a basic for loop when possible, to get rid of an unnecessary Iterator allocation. I also added a unit test for a method I broke out.


stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
```diff
@@ -253,9 +253,16 @@ static boolean isArray(@Nullable TypeMirror type) {
             return true;
         }
         String outerClassType = TypeUtils.getOuterClassType(type);
+        return isSupportedList(type) || outerClassType.equals(Collection.class.getName());
```
kevinzetterstrom: Maybe ```isArray``` should change to ```isSupportedCollection```?

anthonycr: good idea

PR #33, opened 2016-12-22
Suppress unchecked assignments in generated code

mcumings:
The 1.2.0 release generates code with unchecked assignments.  This PR simply adds
@SuppressWarnings("unchecked") to the two locations needed and changes the "sample"
project build config to treat (unsuppressed) unchecked assignments as errors.

This allows the library to be used in projects (like mine) which fail on linter warnings.


anthonycr:
👍 those warnings can be annoying

Issue #27, opened 2016-12-08
Support for a class level annotation


yasirmhd:
#### Issue Summary
Now that Stag has support for Generics, Enums, Maps etc it can generate a type adapter for pretty much any class. With the current approach of having to provide GsonAdapterKey is a huge friction for people to move to this library from Gson.

We can have another annotation maybe at a class level called @UseStag. We have some sort of implementation for this already in place.

anthonycr:
Great idea, such an annotation could maybe replace the current usage of the `@GsonAdapterKey` annotation at the class level (currently used to tell stag about a class that doesn't have any annotated fields but inherits from a class with annotated fields).

This should be simple to implement. 

Let's solidify the spec:
- name the new annotation `@UseStag`
- annotation is only applicable at the class level
- annotation replaces current usage of `@GsonAdapterKey` at the class level
- `@GsonAdapterKey` will no longer be allowed to be applied at the class level
- annotation will cause stag to generate a type adapter with all the fields in the annotated class, regardless of whether or not they are annotated with `@GsonAdapterKey`
- fields in a class annotated with `@UseStag` that are **also** annotated with `@GsonAdapterKey` will use the value passed by that annotation for the json key rather than the field name.

Look good or should any changes be made to this spec? Possibly, the `@GsonAdapterKey` behavior at the class level should be left, in case the consumer wants to generate a type adapter for a class inheriting from another class, but doesn't want any of the fields in the annotated class filled. Or maybe there could be an optional boolean that could be passed to `@UseStag` that specified whether only the parent's fields should be considered

yasirmhd:
Sounds good. The only slight addition that I will like to propose is to give support for SerializedName for JSON name

PR #35, opened 2016-12-23
Stag v2 : Better Performance, Extensible, Class level Annotation, Code de-duplication and Multi-Module Support

anirudhramanan:
**Overview :** 

This change reduces the number of TypeTokens almost to zero. This also handles initialising known Generic classes adapters without using reflection. It also handles the scenarios which were promised during the review of https://github.com/vimeo/stag-java/pull/26 (nested maps and lists do not require TypeTokens anymore). It also fixes https://github.com/vimeo/stag-java/issues/27

**Features Added :**

- **KnownTypeAdapters** : 

  Added a class in the stag library which contains list of all known type adapters such as for Integer, JsonObject, JsonArray, int[], Object types etc.

  It also contains MapTypeAdapter and ListTypeAdapter which takes in the key and valueTypeAdapter in its constructor for reading and writing the keys and values respectively. This change adds support for nested maps, and nested lists, where the valueTypeAdapter will again be a MapTypeAdapter or ListTypeAdapter in those scenarios.

- **Generic Support for TypeAdapters** : 

  **Problem:**

  For scenarios where we have a parameterized generic class, we used to create a constructor in its TypeAdapater which used to take the typeadapter for the parameterized type.

  At the same time, in the Stag Factory file, the getter method for the same generic class used to get the adapter via TypeToken which was expensive, and also did not made sense since we already had the type adapter for that class.

  **Solution:**

  With this change, now in the Stag Factory , instead of creating the typeToken, the object of TypeAdapter will be created, and the type adapter for the parameterized type will be passed in the constructor. For cases, where we do not have the type adapter for its type or its class, as a last resort, it will be reverted back to the TypeToken code.

- **Class Level Annotation Support**:

  Added class level annotations `@UseStag` which will process all the member variables of the class that has been annotated. This will ease the process of integrating Stag.

  We also have `@GsonAdapterKey` support which can now only be applied to the fields, and not at class level. For JSON names, if the field is annotated with both `@GsonAdapterKey` and `@SerializedName`, `@GsonAdapterKey` will get the first preference.

- **Support for Different Maps  and List SubTypes**:

  With the introduction of KnownTypeAdapters, we now support different types of maps such as HashMap, LinkedHashMap, ConcurrentHashMap or ArrayMap. The MapTypeAdapter also expects a ObjectConstructor as a third param, which will return the type of Map to be instantiated. Similarly, the ListTypeAdapter expects ObjectConstructor as a second param which will return the type of the List to be instantiated.

- **Support for Primitive Array Type Parsing**:

  Now Stag will support parsing of int[], byte[], short[], long[], double[], Integer[], String[] and similar types.

- **Support for JsonElement**:

  TypeAdapter support for JsonElement and its subtypes such as JsonObject, and JsonArray

- **Bringing the parsing of primitive types closer to Gson implementation**:

  Gson is able to parse a field as string if the data over the wire is coming as boolean, string or one of the numbers (long, double, int). Likewise if a field is an Integer and over the wire it is coming as String it is able to parse it as integer. With this change it is at absolute parity with Gson. This once again will make the transition of apps/libraries moving from Gson to Stag easier and bug free.

- **Reduced the code duplication**:

  The amount and size of the generated code should also improve given that for the known types we now have pre-defined adapters. This also makes the code more extensible as we can keep adding more adapters (for Dates, SpareArray, etc) without having to change the code generation logic.

- **Do not serialize/deserialize fields that are marked as transient**:


yasirmhd:
@anthonycr  : Kindly review this.


Next, we should probably pick up profiling Stag Vs other popular libraries. We can fork from this repository https://github.com/cowtowncoder/java-json-performance-benchmarks and add the benchmarks for Stag.

anthonycr:
Gonna get to this tomorrow 👍

stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/KnownTypeAdapterUtils.java
```diff
@@ -0,0 +1,168 @@
...
+    /**
+     * Get the instantiater for native array types
+     *
+     * @param typeMirror TypeMirror typeMirror
+     * @return instantiater
+     */
+    @Nullable
+    public static String getNativeArrayInstantiater(@NotNull TypeMirror typeMirror) {
+        String outerClassType = TypeUtils.getOuterClassType(typeMirror);
+        return "new com.vimeo.stag.KnownTypeAdapters.PrimitiveArrayConstructor<" + outerClassType + ">(){ @Override public " + outerClassType + "[] construct(int size){ return new " + outerClassType + "[size]; } }";
```
anthonycr: 👍

stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/TypeUtils.java
```diff
@@ -170,6 +208,26 @@ public static boolean isConcreteType(@NotNull TypeMirror typeMirror) {
         return true;
     }
 
+
+    public static boolean containsTypeVarParams(@NotNull TypeMirror typeMirror) {
```
anthonycr: could we get some javadocs on this method, as it would be the only public method without docs?

stag-library/src/main/java/com/vimeo/stag/UseStag.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Use this annotation to tell Stag that all the member variables of the class should be processed.
+ * <p/>
+ * If the class is annotated, Stag will generate a TypeAdapter for that class.
+ * Stag does not generate TypeAdapters for abstract classes.
+ * <p/>
+ */
+@Target({ElementType.TYPE})
+public @interface UseStag {
```
mcumings: Relying on `transient` to prevent serialization would be incomplete from a Gson perspective as it does not take into account models which leverage `@Expose` and the corresponding `GsonBuilder#excludeFieldsWithoutExposeAnnotation` configuration.  Since the models cannot have compile-time knowledge of the GsonBuilder configuration I'm thinking it's a non-starter to drop `GsonAdapterKey` as proposed.

Given the above, it should probably be an error condition to have both `@GsonAdapterKey` and `@UseStag` in use on a single class.

sample/src/main/java/com/vimeo/sample/model/AccessModifiers.java
```diff
@@ -24,23 +24,28 @@
...
+    @GsonAdapterKey("protectedModifier")
     protected String protectedModifier;
 
-    @GsonAdapterKey
-    public String publicModifier;
+    @GsonAdapterKey("defaultModifier")
+    String defaultModifier;
+
+    //do not include fields of these types in TypeAdapter
+    public static final String DEMO = "demo";
```
anirudhramanan: Agreed

sample-model/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,3 @@
+<resources>
+    <string name="app_name">sample-model</string>
+</resources>
```
anirudhramanan: We'll clean up the sample-model, and remove the files that are not required

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
```diff
@@ -62,141 +61,21 @@
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.type.TypeVariable;
 
-@SuppressWarnings("StringConcatenationMissingWhitespace")
 public class TypeAdapterGenerator extends AdapterGenerator {
 
     private static final String TYPE_ADAPTER_FIELD_PREFIX = "mTypeAdapter";
+    private static boolean mGsonVariableUsed;
+    private static boolean mStagFactoryUsed;
```
anirudhramanan: Agreed (y)

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
```diff
@@ -0,0 +1,795 @@
...
+        }
+    }.nullSafe();
+
+    public static final TypeAdapter<ArrayList<Integer>> INTEGER_ARRAY_LIST_ADAPTER = new ListTypeAdapter<>(INTEGER, new ArrayListInstantiater<Integer>());
+    public static final TypeAdapter<ArrayList<Long>> LONG_ARRAY_LIST_ADAPTER = new ListTypeAdapter<>(LONG, new ArrayListInstantiater<Long>());
+    public static final TypeAdapter<ArrayList<Double>> DOUBLE_ARRAY_LIST_ADAPTER = new ListTypeAdapter<>(DOUBLE, new ArrayListInstantiater<Double>());
+    public static final TypeAdapter<ArrayList<Short>> SHORT_ARRAY_LIST_ADAPTER = new ListTypeAdapter<>(SHORT, new ArrayListInstantiater<Short>());
+    public static final TypeAdapter<ArrayList<Float>> FLOAT_ARRAY_LIST_ADAPTER = new ListTypeAdapter<>(FLOAT, new ArrayListInstantiater<Float>());
+    public static final TypeAdapter<ArrayList<Boolean>> BOOLEAN_ARRAY_LIST_ADAPTER = new ListTypeAdapter<>(TypeAdapters.BOOLEAN, new ArrayListInstantiater<Boolean>());
+    public static final TypeAdapter<ArrayList<Byte>> BYTE_ARRAY_LIST_ADAPTER = new ListTypeAdapter<>(BYTE, new ArrayListInstantiater<Byte>());
```
anirudhramanan: It was kept public so that you can access it directly from outside.

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/model/AnnotatedClass.java
```diff
@@ -51,16 +50,62 @@
...
+                    variableElement.getEnclosingElement().asType() +
+                    ", field must not be private.");
+        }
+    }
+
+    private void addToSupportedTypes(@NotNull Element element) {
+        if (element instanceof VariableElement) {
+            final VariableElement variableElement = (VariableElement) element;
+            Set<Modifier> modifiers = variableElement.getModifiers();
+            if ((!modifiers.contains(Modifier.FINAL) || !modifiers.contains(Modifier.STATIC)) && !modifiers.contains(Modifier.TRANSIENT)) {
```
anirudhramanan: Yes. We'll update the readme doc with all the changes/features that have been integrated

stag-library/src/main/java/com/vimeo/stag/UseStag.java
anthonycr: I hadn't given thought to that GsonBuilder method (or even the exclude with modifiers one), you're right, just relying on transient would be a problem. Maybe we should introduce our own "excludeFieldsWith...?"

I also forgot about the use case for `@GsonAdapterKey` where you can annotate a class without any fields (the example in the model being `VideoList`). In that case we don't want any direct fields of that class touched by stag, but we want all the inherited fields to be resolved and parsed/serialized. `@UseStag` currently would just grab all the fields rather than just tagging the class for stag, which would be problematic.

As to your last comment, that seems like a possibility - require either `@GsonAdapterKey` + `@SerializedName` usage, or `@UseStag` + `@SerializedName`. My fear with enforcing this is that it would become very confusing for consumers of the library as the nuances between the use of the two annotations are confusing without in depth use of the library.

Another option would be for `@UseStag` to take on the responsibilities of `@GsonAdapterKey`. When applied at the class level, it would specify whether stag should pick up all fields or look for annotated fields.

```java
// Use case 1
@UseStag(options = Options.ALL)
public class Model {
    transient String ignored;

    String string1;

    @SerializedName("stringy")
    String string2;

    @UseStag("strings")
    List<String> stringList;
}

// Use case 2
@UseStag(options = Options.NONE)
public class Model2 extends Model3<String> {
    boolean initializationFlag;

    // If this field exists, we'll pick it up.
    // If it doesn't exist, we'll just generate a type adapter based on the inherited, resolved fields.
    @SerializedName("optional")
    boolean optionalThing;
}

public class Model3<T> {
    @UseStag
    T data;
}

// Use case 3
public class Model4 {
    String notSerialized;

    @UseStag
    String serialized;

    @UseStag("serialized_data")
    String serialized2;
}
```

Or we could use `@GsonAdapterKey` instead of the `@UseStag` annotation since that's what consumers are already using. However, since this will be for a 2.0 release, a name change shouldn't be a problem.

cc @kvenn @kevinzetterstrom

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
```diff
@@ -31,6 +31,8 @@
 import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonToken;
 import com.google.gson.stream.JsonWriter;
+import com.sun.istack.internal.NotNull;
+import com.sun.istack.internal.Nullable;
```
anthonycr: Should be using
```java
import org.jetbrains.annotations.NotNull;
import org.jetbrains.annotations.Nullable;
```
instead

stag-library/src/main/java/com/vimeo/stag/UseStag.java
kevinzetterstrom: I am a little confused on why there is the need to add a ```@UseStag``` annotation. The current ```@GsonAdapterKey``` annotation can be applied to the class level and to member variables. In the PR description, it mentions that ```UseStag``` can help in the ease of integration so that consumers need only annotate a model, but that is already in place.

Perhaps ```@UseStag``` can replace the class level ```@GsonAdapterKey``` while the latter is kept in place on a variable level? Although they would perform the same thing, it may be more clear to use.

Having said that, if a new annotation does not bring additional and/or different benefits than the current annotation, then I would be slightly opposed to a name change just for the sake of a name change. From a consumption standpoint I hate when library releases make you refactor all of your code. I would prefer to keep my code in tact but reap in the benefits from library updates - even major updates. Considering this library is still in its infancy yet has many adopters, I would personally prefer that we stick with the current annotation name, if possible.

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
anirudhramanan: done

sample/src/main/java/com/vimeo/sample/model/AccessModifiers.java
anirudhramanan: done

sample-model/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,9 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.vimeo.sample_model">
+
+    <application android:allowBackup="true" android:label="@string/app_name"
+        android:supportsRtl="true">
+
+    </application>
+
+</manifest>
```
anirudhramanan: done

stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/KnownTypeAdapterUtils.java
```diff
@@ -0,0 +1,168 @@
...
+        if (outerClassType.equals(Map.class.getName())) {
+            return "new com.vimeo.stag.KnownTypeAdapters.MapInstantiater" + postFix;
+        } else if (outerClassType.equals(HashMap.class.getName())) {
+            return "new com.vimeo.stag.KnownTypeAdapters.HashMapInstantiater" + postFix;
+        } else if (outerClassType.equals(LinkedHashMap.class.getName())) {
+            return "new com.vimeo.stag.KnownTypeAdapters.LinkedHashMapInstantiater" + postFix;
+        } else if (outerClassType.equals(ConcurrentHashMap.class.getName())) {
+            return "new com.vimeo.stag.KnownTypeAdapters.ConcurrentHashMapInstantiater" + postFix;
+        } else {
+            return "new com.google.gson.internal.ObjectConstructor<" + outerClassType + "<" + keyType.toString() + ", " + paramType.toString() + ">>() " +
```
anirudhramanan: done

stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/KnownTypeAdapterUtils.java
```diff
@@ -0,0 +1,168 @@
...
+     * @param typeMirror TypeMirror typeMirror
+     * @return instantiater
+     */
+    @NotNull
+    public static String getMapInstantiater(@NotNull TypeMirror typeMirror) {
+        String outerClassType = TypeUtils.getOuterClassType(typeMirror);
+        DeclaredType declaredType = typeMirror instanceof DeclaredType ? (DeclaredType) typeMirror : null;
+        TypeMirror keyType = declaredType != null && declaredType.getTypeArguments() != null && declaredType.getTypeArguments().size() == 2 ? declaredType.getTypeArguments().get(0) : null;
+        TypeMirror paramType = declaredType != null && declaredType.getTypeArguments() != null && declaredType.getTypeArguments().size() == 2 ? declaredType.getTypeArguments().get(1) : null;
+        String postFix = keyType != null ? "<" + keyType.toString() + ", " + paramType.toString() + ">()" : "()";
```
anirudhramanan: done

stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/KnownTypeAdapterUtils.java
```diff
@@ -0,0 +1,168 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicIntegerArray;
+
+import javax.lang.model.type.DeclaredType;
+import javax.lang.model.type.TypeMirror;
+
+/**
+ * This maintains a list of type vs the known type adapters.
+ */
+public class KnownTypeAdapterUtils {
```
anirudhramanan: done

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
```diff
@@ -236,95 +113,432 @@ private static String getTypeTokenCode(@NotNull TypeMirror fieldType,
...
+            }
+        }
+        return result;
     }
 
     @NotNull
-    private static MethodSpec getWriteMethodSpec(@NotNull TypeName typeName,
-                                                 @NotNull Map<Element, TypeMirror> memberVariables,
-                                                 @NotNull AdapterFieldInfo adapterFieldInfo) {
+    private MethodSpec getWriteMethodSpec(@NotNull TypeName typeName, @NotNull Map<Element, TypeMirror> memberVariables, @NotNull AdapterFieldInfo adapterFieldInfo) {
```
anirudhramanan: done

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
```diff
@@ -236,95 +113,432 @@ private static String getTypeTokenCode(@NotNull TypeMirror fieldType,
...
+            mGsonVariableUsed = true;
+            mStagFactoryUsed = true;
+            return "mStagFactory." + "get" + getterField + "(mGson)";
         }
     }
 
+    /**
+     * Returns the adapter code for the unknown types.
+     */
+    private String getAdapterForUnknownType(@NotNull TypeMirror fieldType, @NotNull TypeSpec.Builder adapterBuilder,
```
anirudhramanan: done

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
```diff
@@ -236,95 +113,432 @@ private static String getTypeTokenCode(@NotNull TypeMirror fieldType,
...
+            if (fieldType.getKind() == TypeKind.TYPEVAR) {
+                result = typeVarsMap.get(fieldType);
+            } else if (fieldType instanceof DeclaredType) {
+                DeclaredType declaredFieldType = (DeclaredType) fieldType;
+                List<? extends TypeMirror> typeMirrors = ((DeclaredType) fieldType).getTypeArguments();
 
-    private static boolean isNativeArray(@NotNull TypeMirror type) {
-        return (type instanceof ArrayType);
-    }
+                if (null == typeMirrors || typeMirrors.size() == 0) {
```
anirudhramanan: done

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
anirudhramanan: done

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/ExternalAdapterInfo.java
```diff
@@ -0,0 +1,100 @@
...
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.Name;
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.type.DeclaredType;
+import javax.lang.model.type.TypeMirror;
+
+public class ExternalAdapterInfo {
+
+    @NotNull
+    private static Set<String> sCheckedClasses = new HashSet<>();
```
anirudhramanan: done

sample/src/main/java/com/vimeo/sample/model/PrimitiveTypesExample.java
```diff
@@ -0,0 +1,27 @@
...
+ * Model class which contains different types of primitive member variables
+ */
+@UseStag
+public class PrimitiveTypesExample {
+
+    @GsonAdapterKey("simpleInt")
+    public int simpleInt;
+
+    @GsonAdapterKey("simpleFlot")
+    public float simpleFlot;
```
anirudhramanan: done

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/StagGenerator.java
```diff
@@ -264,31 +420,90 @@ private TypeSpec getAdapterFactorySpec() {
...
-            result = fieldNameBuilder.toString();
-            mUnknownClasses.add(classInfo);
-            mUnknownAdapterFieldMap.put(fieldTypeString, result);
         }
+
         return result;
     }
+
+    @NotNull
+    private String removeSpecialCharacters(TypeMirror typeMirror) {
```
anirudhramanan: done

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/StagGenerator.java
```diff
@@ -264,31 +420,90 @@ private TypeSpec getAdapterFactorySpec() {
...
+                result = removeSpecialCharacters(typeMirror);
+                return result + "$$PrimitiveArray" + "$$";
+            } else if (typeMirror instanceof DeclaredType) {
+                List<? extends TypeMirror> typeArguments = ((DeclaredType) typeMirror).getTypeArguments();
+                if (typeArguments == null || typeArguments.size() == 0) {
+                    result = removeSpecialCharacters(typeMirror) + "$$";
                 } else {
-                    fieldNameBuilder.append(makeCapital ? Character.toUpperCase(c) : c);
-                    makeCapital = false;
+                    result += outerClassType + "$$";
```
anirudhramanan: done

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/StagGenerator.java
```diff
@@ -264,31 +420,90 @@ private TypeSpec getAdapterFactorySpec() {
...
+    private String generateMethodName(@NotNull TypeMirror typeMirror) {
+        String result = "";
+        String outerClassType = TypeUtils.getSimpleOuterClassType(typeMirror);
+        if (TypeUtils.isConcreteType(typeMirror)) {
+            if (TypeUtils.isNativeArray(typeMirror)) {
+                result = removeSpecialCharacters(typeMirror);
+                return result + "$$PrimitiveArray" + "$$";
+            } else if (typeMirror instanceof DeclaredType) {
+                List<? extends TypeMirror> typeArguments = ((DeclaredType) typeMirror).getTypeArguments();
+                if (typeArguments == null || typeArguments.size() == 0) {
```
anirudhramanan: done

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/StagGenerator.java
```diff
@@ -36,72 +36,196 @@
...
+                if (!checkKnownAdapters(type)) {
+                    hasUnknownTypeFields = true;
+                    break;
                 }
-                knownFieldNames.add(adapterFactoryMethodName);
             }
-            mKnownClasses.add(classInfo);
-            mFieldNameMap.put(knownType.toString(), adapterFactoryMethodName);
+
+            mGenericClassInfo.put(knownGenericType.getType().toString(), new GenericClassInfo(typeArguments.size(), hasUnknownTypeFields));
```
anirudhramanan: done

sample/src/main/java/com/vimeo/sample/model/ExternalModelExample.java
```diff
@@ -0,0 +1,40 @@
...
+    @SerializedName("externalExample")
+    public ExternalModel1 mExternalModel;
+
+    @SerializedName("externalGenericExample")
+    public ExternalModelGeneric<String> mExternalGenericExample;
+
+    @SerializedName("parametrizedExternalGenericExample")
+    public ExternalModelGeneric<T> mParametrizedExternalGenericExample;
+
+
```
anirudhramanan: done

stag-library-compiler/src/test/java/com/vimeo/stag/processor/generators/TypeAdapterGeneratorUnitTest.java
```diff
@@ -53,154 +53,113 @@
     public void testIsSupportedPrimitive_supportsTypes() throws Exception {
```
anirudhramanan: done

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
```diff
@@ -0,0 +1,795 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import static com.google.gson.stream.JsonToken.BEGIN_ARRAY;
```
anirudhramanan: done

sample-model/build.gradle
```diff
@@ -0,0 +1,51 @@
...
+    compileSdkVersion 23
+    buildToolsVersion "23.0.3"
+
+    defaultConfig {
+        minSdkVersion 14
+        targetSdkVersion 23
+        versionCode 1
+        versionName "1.0"
+
+        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
```
anirudhramanan: done

sample-model/build.gradle
```diff
@@ -0,0 +1,51 @@
...
+    }
+}
+
+dependencies {
+    compile fileTree(dir: 'libs', include: ['*.jar'])
+    androidTestCompile('com.android.support.test.espresso:espresso-core:2.2.2', {
+        exclude group: 'com.android.support', module: 'support-annotations'
+    })
+    compile 'com.android.support:appcompat-v7:23.4.0'
+    testCompile 'junit:junit:4.12'
```
anirudhramanan: done

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
```diff
@@ -0,0 +1,795 @@
...
+ * are used to read and write keys and values respectively, and the ObjectConstructor tells the type of Map to be instantiated. This will also support the scenario
+ * where we have a nested maps. In that case the valueTypeAdapter will be again a {@link MapTypeAdapter} with its key and value TypeAdapters
+ * <p>
+ * --------------------- {@link ListTypeAdapter} -------------------
+ * <p>
+ * In case of {@link ListTypeAdapter}, we need to pass the valueTypeAdapter and the ObjectConstructor. The valueTypeAdapter
+ * is used to read and write the values, and the ObjectConstructor tells the type of List to be instantiated. This will also support the scenario
+ * where we have a nested list. In that case the valueTypeAdapter will be again a {@link ListTypeAdapter} with its value TypeAdapter
+ */
+public class KnownTypeAdapters {
```
anirudhramanan: done

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
```diff
@@ -0,0 +1,795 @@
...
+            T[] result = this.mObjectCreator.construct(object.size());
+            return object.toArray(result);
+        }
+
+    }
+
+    /**
+     * Type Adapter for Integer[] type. This can be directly accessed to read and write
+     */
+    public static final class PrimitiveIntegerArrayAdapter {
```
anirudhramanan: done

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
```diff
@@ -0,0 +1,795 @@
...
+        public static void write(JsonWriter writer, int[] value) throws IOException {
```
anirudhramanan: done

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
```diff
@@ -0,0 +1,795 @@
...
+            if (null != value) {
+                writer.beginArray();
+                for (int item : value) {
+                    writer.value(item);
+                }
+                writer.endArray();
+            }
+        }
+
+        public static int[] read(JsonReader reader) throws IOException {
```
anirudhramanan: done

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
```diff
@@ -0,0 +1,795 @@
...
+                writer.beginArray();
+                for (char item : value) {
+                    writer.value(item);
+                }
+                writer.endArray();
+            }
+        }
+
+        public static char[] read(JsonReader reader) throws IOException {
+            String string = com.google.gson.internal.bind.TypeAdapters.STRING.nullSafe().read(reader);
```
anirudhramanan: done

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
```diff
@@ -0,0 +1,795 @@
...
+            return null != string ? string.toCharArray() : null;
+        }
+    }
+
+    /**
+     * Default Instantiater for List, by default it will create the Map of {@link ArrayList} type
+     */
+    public static class ListInstantiater<V> implements ObjectConstructor<List<V>> {
```
anirudhramanan: done

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
```diff
@@ -0,0 +1,795 @@
...
+        }
+    }
+
+    /**
+     * Default Instantiater for Maps, by default it will create the Map of {@link LinkedHashMap} type
+     */
+    public static class MapInstantiater<K, V> implements ObjectConstructor<Map<K, V>> {
+        @Override
+        public Map<K, V> construct() {
+            return new LinkedHashMap<K, V>();
```
anirudhramanan: done

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
```diff
@@ -0,0 +1,795 @@
...
+        }
+    }
+
+    /**
+     * Type Adapter for {@link Collection}
+     */
+    public static class ListTypeAdapter<V, T extends Collection<V>> extends TypeAdapter<T> {
+
+        private TypeAdapter<V> valueTypeAdapter;
+        private ObjectConstructor<T> objectConstructor;
```
anirudhramanan: done

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
```diff
@@ -0,0 +1,795 @@
...
+    }
+
+    /**
+     * Type Adapter for {@link Map}. The constructor expects {@link ObjectConstructor} which is
+     * used to instantiate maps of particular types eg, {@link HashMap} {@link LinkedHashMap} etc
+     */
+    public static class MapTypeAdapter<K, V, T extends Map<K, V>> extends TypeAdapter<T> {
+        private ObjectConstructor<T> objectConstructor;
+        private TypeAdapter<V> valueTypeAdapter;
+        private TypeAdapter<K> keyTypeAdapter;
```
anirudhramanan: done

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
```diff
@@ -0,0 +1,795 @@
...
+        public void write(JsonWriter writer, T value) throws IOException {
+            if(null == value) {
+                writer.nullValue();
+                return;
+            }
+            boolean hasComplexKeys = false;
+            List<JsonElement> keys = new ArrayList<>(value.size());
+
+            List<V> values = new ArrayList<>(value.size());
+            for (T.Entry<K, V> entry : value.entrySet()) {
```
anirudhramanan: yes, done

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
```diff
@@ -0,0 +1,795 @@
...
+                    if (replaced != null) {
+                        throw new JsonSyntaxException("duplicate key: " + key);
+                    }
+                }
+                in.endObject();
+            }
+            return map;
+        }
+
+        private String keyToString(JsonElement keyElement) {
```
anirudhramanan: done

stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/TypeUtils.java
anirudhramanan: added doc

stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/FileGenUtils.java
```diff
@@ -43,7 +43,7 @@
 
 public final class FileGenUtils {
 
-    private static final String UNESCAPED_SEPARATOR = "$";
+    public static final String UNESCAPED_SEPARATOR = "$";
```
anirudhramanan: done

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/model/AnnotatedClass.java
```diff
@@ -51,16 +50,62 @@
...
+
+    private static void checkModifiers(VariableElement variableElement, Set<Modifier> modifiers) {
+        if (modifiers.contains(Modifier.FINAL)) {
+            if (!modifiers.contains(Modifier.STATIC)) {
+                throw new RuntimeException("Unable to access field \"" +
+                        variableElement.getSimpleName().toString() + "\" in class " +
+                        variableElement.getEnclosingElement().asType() +
+                        ", field must not be final.");
+            }
+        } else if (modifiers.contains(Modifier.PRIVATE)) {
```
anirudhramanan: done

stag-library/src/main/java/com/vimeo/stag/UseStag.java
yasirmhd: Thanks all for reviewing this. 

I will first begin by sharing the challenges when we were trying to move to Stag. We have more than 350+ model in different classes spread across different modules. Each of these classes have several member variables. We also do use @SerializedName in some of the cases. Putting a GsonAdapterKey to all these member variables was a cumbersome job. If the member variable name and the JSON serialized name was exactly the same then it was turning out an overkill for developers to add an annotation to every field. This is when we thought we add an annotation at a class level since STAG generates adapters and adapters are at a class level. So if one had to exclude a particular class they just do not add that Annotation to the class.

While raising this PR we did have the early adopters in mind, and hence we did retain the GsonAdapterKey for that backward compatibility. Existing STAG users will have to add an annotation at a class level to migrate to the never version - and hence the version bump up. 

As @kevinzetterstrom rightly mentions, keeping separate annotations for class level and field level was to have clarity in usage. Also like @anthonycr 's recommendation to add Options.ALL, Options.NONE we kept it separate as the class level specifications are always going to be different from the field level ones (Options.IGNORE_BASE_CLASS_FIELDS etc). It makes it more flexible in the future and we will not have to make breaking changes in future if we keep them separate.

To alleviate the pains of folks moving from v1 to v2 we can support GsonAdapterKey Or UseStag both - but that will end up in a code mess for people where half of the code is doing it in one way and the other half in a different way. Also for the library it will be an overhead to keep supporting both the ways.


@mcumings  : We will like to provide @Expose functionality at some point in time as well. When implemented the generate code will have an if-else condition which checks the field being exposed at run time. The if else clause will exist only for the fields which have @Expose annotation set. Also on the lines of what @anthonycr mentions we can have compile time exclusion policies in a future release.

If none of these works here is an alternative that should hopefully work for all - 

- Support GsonAdapterKey both at class and field level
- If this comes at the class level all member variables automatically are considered for parsing (irrespective of whether they have or do not have the GsonAdapterKey)
- For excluding only a few members from being serialized/deserialized do not include the GsonAdapterKey at the class level, but include it for the fields which need to be serialized/deserialized (not the ideal approach once again since it is only one or two fields which need to be excluded from the Serialization/Deserialization)

yasirmhd: For the transient field related change : As per gson's documentation they do not do this for static and transient fields. We are just ensuring the same

image

yasirmhd: Another thing : Class based annotation will help reduce the build time for STAG processing. 

Btw, that is one of our next task. Optimize the annotation processing and code generation logic to minimise the build time : becomes very important for developer productivity

yasirmhd: @anthonycr , @kevinzetterstrom , @mcumings kindly let us know of your thoughts here. We will make the required changes and update the PR

mcumings: I'm completely happy with a deferred support of `@Expose` and was primarily bringing it up as a delta from Gson behavior, which you are obviously pursuing.  👍  from this consumer.

mcumings:
It looks like this work was based off of the 1.1.2 codebase and will have quite a few merge conflicts.
My project definitely needs the @SuppressWarnings("unchecked") additions and relevant
sample project build script additions.

yasirmhd:
@mcumings We will look into this and fix this

yasirmhd:
@mcumings : The changes in https://github.com/vimeo/stag-java/pull/33/files seem to be all there. Can you please check where exactly do we need to add this suppression? We will add this to the new module for sure

image

mcumings:
derp. my bad, I checked out fk-stag/v2 and not stag/v2.  Disregard...

stag-library/src/main/java/com/vimeo/stag/UseStag.java
anthonycr: After some discussion, this is what we would like to do for the 2.0 release. The ultimate goal would be to make using stag a one line change per model (which is what I think you are trying to drive toward):

1. Keep `@UseStag` as a class level annotation. Add options to allow all fields (excluding static and transient) in the class to be serialized (`FieldOptions.ALL`), or only those marked with `@SerializedName` (`FieldOptions.SERIALIZED_NAME`) or the deprecated `@GsonAdapterKey`

    - default `FieldOptions` will be `FieldOptions.ALL`. 
    - We will require `@UseStag` annotation in order for stag to generate a type adapter.
    - Eventually we can add more field options, but in order to reach parity with 1.x, we will just start out with the above mentioned two field options.
2. Deprecate (but keep around until a 2.1 release) `@GsonAdapterKey`
    - This will encourage moving back to `@SerializedName`. 

Do you think this works @yasirmhd @anirudhramanan @mcumings ? or do you have any name change suggestions for the field options? We can merge this PR as is and make the two changes after the merge, or if you want to make the changes in this PR that's cool too.

mcumings:
After - ahem - pulling the correct branch I'm getting 3 unit test failures in 
`com.vimeo.stag.processor.TypeUtilsUnitTest`.  Looks like the tests might have errors in
them.  e.g., `testIsSupportedList_supportsTypes` is calling `TypeUtils.isSupportedCollection`
instead of the presumably correct `TypeUtils.isSupportedList` method.  Might want to take a
quick look before merging.

anthonycr:
Hmm, yes, seeing the same thing now. I checked them earlier in the review and they were passing. Something changed

yasirmhd:
Will fix the tests in a bit

stag-library/src/main/java/com/vimeo/stag/UseStag.java
yasirmhd: Sounds great. I will update the PR in another 45-60 minutes with this implemented.

mcumings:
Okay, one more issue discovered: A StackOverflow during compilation, as follows:

```
Caused by: java.lang.StackOverflowError
        at com.sun.tools.javac.util.Name.isEmpty(Name.java:104)
        at com.sun.tools.javac.code.Type$ClassType.className(Type.java:823)
        at com.sun.tools.javac.code.Type$ClassType.toString(Type.java:812)
        at com.vimeo.stag.processor.utils.TypeUtils.getInheritedType(TypeUtils.java:247)
        at com.vimeo.stag.processor.generators.model.AnnotatedClass.<init>(AnnotatedClass.java:58)
        at com.vimeo.stag.processor.generators.model.SupportedTypesModel.getSupportedType(SupportedTypesModel.java:93)
        at com.vimeo.stag.processor.generators.model.AnnotatedClass.<init>(AnnotatedClass.java:68)
        at com.vimeo.stag.processor.generators.model.SupportedTypesModel.getSupportedType(SupportedTypesModel.java:93)
        at com.vimeo.stag.processor.generators.model.AnnotatedClass.addToSupportedTypes(AnnotatedClass.java:107)
        at com.vimeo.stag.processor.generators.model.AnnotatedClass.<init>(AnnotatedClass.java:61)
        at com.vimeo.stag.processor.generators.model.SupportedTypesModel.getSupportedType(SupportedTypesModel.java:93)
        at com.vimeo.stag.processor.generators.model.AnnotatedClass.<init>(AnnotatedClass.java:68)
        at com.vimeo.stag.processor.generators.model.SupportedTypesModel.getSupportedType(SupportedTypesModel.java:93)
        at com.vimeo.stag.processor.generators.model.AnnotatedClass.addToSupportedTypes(AnnotatedClass.java:107)
        at com.vimeo.stag.processor.generators.model.AnnotatedClass.<init>(AnnotatedClass.java:61)
        at com.vimeo.stag.processor.generators.model.SupportedTypesModel.getSupportedType(SupportedTypesModel.java:93)
        at com.vimeo.stag.processor.generators.model.AnnotatedClass.<init>(AnnotatedClass.java:68)
        ...
```

yasirmhd:
@mcumings  can you please the class for which this is happening?

mcumings:
Unfortunately, the error is associated with a file that is unrelated (a dagger 2 file) because that
annotation processor didn't run and generate code.

I'll try to get more information.

mcumings:
I was able to reduce the issue to the following structure:
```
@UseStag
public class NestedClass {
    public static class Nested extends NestedClass {}
}
```

yasirmhd:
@mcumings will fix this in the next PR. Another 20 minutes or so

stag-library/src/main/java/com/vimeo/stag/UseStag.java
yasirmhd: done. 

@anthonycr can you please review

yasirmhd:
@mcumings Fixed StackOverflow. Very good catch. Thank you. Can you please verify.

stag-library-compiler/src/test/java/com/vimeo/stag/processor/TypeUtilsUnitTest.java
```diff
@@ -315,11 +315,11 @@ public void testIsSupportedPrimitive_supportsTypes() throws Exception {
         Assert.assertTrue(TypeUtils.isSupportedPrimitive(boolean.class.getName()));
```
anthonycr: 👍

mcumings:
StackOverflow fix confirmed, thanks!

I've found one more issue around List / ArrayList types that I'm working to isolate...

mcumings:
Okay, I'm getting pulled away for a few hours and haven't yet been able to isolate the scenario.  If you want to try to work backwards, I'm getting a compile-time error of:
```
Error:(47, 134) error: incompatible types: List<Parameter> cannot be converted to ArrayList<Parameter>
```

for the TypeAdapter.write code:
```
    if (object.parameters != null)  {
      writer.name("parameters");
      mStagFactory.getArrayList$cErrorLongDetailsParameter$TypeAdapter(mGson).write(writer, object.parameters);
    }
```

The field is defined within a convoluted model class hierarchy as follows:
```
public List<Parameter> parameters;
```

But is is also used as follows elsewhere as:
```
public ArrayList<Parameter> parameters;
```

That's as far as I've gotten thus far...

stag-library/src/main/java/com/vimeo/stag/UseStag.java
anthonycr: @yasirmhd the changes look great, thank you

yasirmhd:
@mcumings Not able to reproduce this one. We will keep trying to see if we can hit this scenario

yasirmhd:
@mcumings do you have a class with Parameter defined in multiple packages? What about the usage of Parameter in difference classes?

`      mStagFactory.getArrayList$cErrorLongDetailsParameter$TypeAdapter(mGson).write(writer, object.parameters);
`

Did you change this line to obfuscate the package name?

cErrorLong seems wrong.

mcumings:
I did obfuscate the name, yes.  It was also insanely long due to the codebase in question.

Yes, I believe the Parameter class is used from multiple packages.  That's a great insight.

yasirmhd:
@mcumings : No luck so far. We have tried all combinations and have not been able to reproduce this

anirudhramanan:
@mcumings The method name `mStagFactory.getArrayList$cErrorLongDetailsParameter$TypeAdapter(mGson).write(writer, object.parameters);` seems wrong. Since in the code path, even for the conflicting names that are used inside Collections or Map, the package name should be in smaller case.

yasirmhd:
@mcumings : I have a repro for this one - finally

mcumings:
Got out of my meeting early and was able to isolate this.  This goes way off the reservation
of good design but it is what is in my code base.  It should probably be handled but part of me
would be happy to have you reject that notion. 😄 

```
@UseStag
public class ListScenario {
    public ArrayList<Stats> parameters;
}
```

and the killer extension class which overrides the field type (grunt):
```
@UseStag
public class ListScenarioFieldTypeOverride extends ListScenario {
    public List<Stats> parameters;
}
```

yasirmhd:
@anthonycr , @mcumings : fixed :-) btw, Gson would throw a run time exception in this case ;-)

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/model/AnnotatedClass.java
```diff
@@ -62,24 +63,38 @@
...
+
+            for(Map.Entry<Element, TypeMirror> entry : inheritedMemberVariables.entrySet()) {
+                if(!variableNames.contains(entry.getKey().getSimpleName().toString())) {
+                    mMemberVariables.put(entry.getKey(), entry.getValue());
+                } else {
+                    if (StagProcessor.DEBUG) {
+                        DebugLog.log(TAG, "\t\tIgnoring inherited Member variables - " + entry.getKey().asType().toString());
+                    }
+                }
+            }
```
anthonycr: Simple solution, looks and works great!

mcumings:
Alright!  Got a clean compile!  Thanks for fixing those issues!

anthonycr:
I think we are ready to merge this?? Any objections?

anirudhramanan:
Good to go !

anthonycr:
Merged, I'll be working on getting the release out over the next few days: at the earliest by friday, more realistically monday.

mcumings:
Thanks, all!  This is a nice chunk of work!

mcumings:
Any update on the release timeframe, by chance?

anthonycr:
Sorry for the delay, mike, I'm trying to get it out before the end of the week

mcumings:
Okay, thanks for the update.

PR #38, opened 2017-01-13
Parsing using TypeAdapters

anirudhramanan:
Since gson.fromJson("string", .class) internally uses reflection to figure out the type and to call it's corresponding type adapter by iterating through the factory, we can actually avoid this reflection by directly providing the type adapter, instead of asking the gson. 

screen shot 2017-01-13 at 11 24 14 pm

We already have the type adapter for VideoList.class, we can use its type adapter, and serialize/deserialize (this will avoid creation of TypeToken of VideoList.class)

**We ran a profiling (5 iterations) for the same, sharing the results :** 

**Earlier : (avg 13.6 ms)**

Time elapsed while parsing: 12 ms
Time elapsed while parsing: 15 ms
Time elapsed while parsing: 15 ms
Time elapsed while parsing: 13 ms
Time elapsed while parsing: 13 ms

**Now : (avg 10 ms)**

Time elapsed while parsing: 9 ms
Time elapsed while parsing: 10 ms
Time elapsed while parsing: 13 ms
Time elapsed while parsing: 10 ms
Time elapsed while parsing: 8 ms

We were able to shave off around 4-5 ms in each run. Overall, it gave us 26.4 % improvement in the parsing time.

### DIFF
--- a/sample/src/main/java/com/vimeo/sample/network/NetworkRequest.java
+++ b/sample/src/main/java/com/vimeo/sample/network/NetworkRequest.java
@@ -34,10 +34,14 @@ import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
+import com.google.gson.TypeAdapter;
+import com.google.gson.internal.bind.JsonTreeReader;
+import com.google.gson.stream.JsonReader;
 import com.vimeo.sample.model.DateParser;
 import com.vimeo.sample.model.Video;
 import com.vimeo.sample.model.VideoList;
 import com.vimeo.sample.stag.generated.Stag;
+import com.vimeo.stag.KnownTypeAdapters;
 
 import java.io.BufferedReader;
 import java.io.IOException;
@@ -116,16 +120,17 @@ public final class NetworkRequest {
                     builder.append(line);
                 }
 
+                Stag.Factory factory = new Stag.Factory();
                 Gson gson = new GsonBuilder().registerTypeAdapter(Date.class, new DateParser())
-                        .registerTypeAdapterFactory(new Stag.Factory())
+                        .registerTypeAdapterFactory(factory)
                         .create();
 
                 JsonObject jsonObject = new JsonParser().parse(builder.toString()).getAsJsonObject();
 
+                TypeAdapter<VideoList> videoListTypeAdapter = factory.getVideoList$TypeAdapter(gson);
                 long time = System.currentTimeMillis();
 
-                videos.addAll(gson.fromJson(jsonObject, VideoList.class).data);
-
+                videos.addAll(videoListTypeAdapter.fromJsonTree(jsonObject).data);
                 Log.d(TAG, "Time elapsed while parsing: " + (System.currentTimeMillis() - time) + " ms");
 
             } catch (IOException e) {


anirudhramanan:
@anthonycr We came across a very nice [article](http://makingvimeo.com/post/148808044404/boosting-app-performance-with-reflectionless) over stag by vimeo. Probably, we should be running the profiling once again for the same data set.

yasirmhd:
@anthonycr  : Do you have the data set with which you ran the tests mentioned in http://makingvimeo.com/post/148808044404/boosting-app-performance-with-reflectionless ? 

We would like to run another benchmark of these to test where we stand. Also we would like that benchmark to be checked into the code so that we can ensure we are not regressing in terms of performance with any feature that gets added.

anthonycr:
@anirudhramanan @yasirmhd unfortunately I don't think the tests kevin ran for that article were made using a strict data set. I believe he was simply hitting the `https://api.vimeo.com/channels/staffpicks/videos` endpoint that I'm using in stag's sample. 

@kevinzetterstrom were you just hitting the staff picks endpoint or did you use a dummy data set?

anthonycr:
As for this PR, this is a good find. The less reflection we're using, the better 👍 

Also, for a benchmark, I agree we need to find something. I'm going to look and see what viable json benchmarks there are. Otherwise, I could generate a permanent dataset from the vimeo api to test with.

sample/src/main/java/com/vimeo/sample/network/NetworkRequest.java
```diff
@@ -116,16 +120,17 @@ void cancelRequest() {
...
                         .create();
 
                 JsonObject jsonObject = new JsonParser().parse(builder.toString()).getAsJsonObject();
 
+                TypeAdapter<VideoList> videoListTypeAdapter = factory.getVideoList$TypeAdapter(gson);
                 long time = System.currentTimeMillis();
 
-                videos.addAll(gson.fromJson(jsonObject, VideoList.class).data);
-
+                videos.addAll(videoListTypeAdapter.fromJsonTree(jsonObject).data);
```
anthonycr: While this works, do you think it would more helpful if we added an API method to the generated `Stag` class:

```java
class Stag {
    public <T> T fromJson(Gson gson, JsonObject jsonObject, Class<T> clazz) {
        switch (clazz.getName()) {
            case "com.vimeo.sample.model.VideoList":
                return getVideoList$TypeAdapter(gson).fromJsonTree(jsonObject);
        }
    }
}

...
videos.addAll(factory.fromGson(gson, jsonObject, VideoList.class));
```

that way it's the same number of lines as before + the extra line to declare the factory object on its own line. This would be a little cumbersome and redundant to generate though since it would look very similar to the `Factory.create` method.

yasirmhd: This will not be able to handle generic classes. 

This is however a good idea for a future release. We have created a helper class which does something similar : It does not have a switch case though. We can possibly generate a helper class which does the fromJson thing. 

We should do this in a separate class because this class has become too big. For our case where there are 350+model classes the loading of this class is taking 15ms

yasirmhd:
We have removed all usages of TypeToken and .Class in our product and we have seen 35-110% jump in performance. You do that and you can parse json on main threads as well

kevinzetterstrom:
@anirudhramanan @anthonycr I used the ```https://api.vimeo.com/channels/staffpicks/videos``` endpoint, with the addition of several field filters. Since that endpoint is constantly being updated with new data, the data being returned now will be different than the data I used in my tests. While the size of the data shouldn't drastically change, I would hate to not have the same data used each time we want to run a test. I agree with you both, we need to come up with a fixed set of data for which to test against.

yasirmhd:
We can merge this PR and create an issue to have perf benchmarks.

If you can check in sample data set of json files we will write the benchmarks for them

PR #43, opened 2017-01-30
Fix missed model classes when nested within a non-annotated class

mcumings:
#### Ticket Summary

Fixes an issue where model classes were only being detected if the root element was annotated.

Also deletes an example unit test class.

#### Implementation Summary

Switched to a more appropriate means of detecting annotated classes.  Rather than iterating over
root elements and filtering by whether or not they are annotated with `UseStag.class` we use
the `RoundEnvironment#getElementsAnnotatedWith(UseStag.class)`.

#### How to Test

Unit test provided.


mcumings:
A fast turn on this would be appreciated, if possible.

wilrnh:
hey @mcumings I just pushed a fix for the failed check. Would you please pull it in from the latest `dev` branch please?

mcumings:
Done!

anthonycr:
Good catch, and glad to see that unit test 👍 I'll work on getting a 2.0.1 release out today

mcumings:
Thanks, Anthony!

PR #45, opened 2017-01-30
Version 2.0.1

anthonycr:
#### Summary
- Fixed bug with backwards compatibility where classes that contained only `@GsonAdapterKey` annotations were not being picked up by Stag. This backwards compatibility was meant to be in 2.0.0 but was overlooked. This means that the only change needed to upgrade from 1.2.1 to 2.0.1 is that you need to replace class level `@GsonAdapterKey` usage with comparable `@UseStag`.
- Fixed a small bug where inner classes that were not annotated with `@UseStag` but were nested inside classes that were annotated with use stag were auto assigned the `FieldOption.ALL` option. Instead, now it inherits the field option from the outer class as would be expected.
- Added a unit test for the backwards compatibility of generating an adapter for a class with only `@GsonAdapterKey` annotations.
- Updated the version to 2.0.1


sample/src/test/java/com/vimeo/sample/Utils.java
```diff
@@ -0,0 +1,54 @@
...
+ * Utils class for testing type adapter generation.
+ *
+ * Created by restainoa on 1/30/17.
+ */
+public final class Utils {
+
+    private Utils() {
+    }
+
+    private static <T> TypeAdapter<T> getTypeAdapter(@NonNull Class<T> clazz) {
```
kevinzetterstrom: This could use a `Nullable` annotation

PR #48, opened 2017-02-02
Handle the case where element is of type other than a class

anirudhramanan:
For Element Kinds such as @interface , and other element types (apart from CLASS type) , these should not be processed.


anthonycr:
Looks great. Good edge case, merging.

PR #49, opened 2017-02-03
Type adapter compilation safety

anthonycr:
#### Summary
We were using hardcoded strings for all the adapter fields in the `KnownTypeAdaptersUtils` class. This had a couple of effects:
- It made it look like most of the classes and fields in `KnownTypeAdapters` were unused to the compiler. This meant we had to suppress helpful lint warnings in that class.
- It made any refactoring more difficult.
- It made the code susceptible to any other changes, such as changes to Gson. With this change it is much more clear if there is a problem with the class referenced by Gson.

I also did some general refactoring around this code to make it more clear.

### DIFF
--- a/stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/KnownTypeAdapterUtils.java
+++ b/stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/KnownTypeAdapterUtils.java
@@ -40,13 +40,13 @@ public final class KnownTypeAdapterUtils {
     private static final HashMap<String, String> KNOWN_TYPE_ADAPTERS = new HashMap<>();
 
     @NotNull
-    private static final HashMap<String, String> SUPPORTED_COLLECTION_INFO = new HashMap<>();
+    private static final HashMap<String, String> SUPPORTED_COLLECTION_INSTANTIATORS = new HashMap<>();
 
     @NotNull
-    private static final HashMap<String, String> SUPPORTED_MAP_INFO = new HashMap<>();
+    private static final HashMap<String, String> SUPPORTED_MAP_INSTANTIATORS = new HashMap<>();
 
     @NotNull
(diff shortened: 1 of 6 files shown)


stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/KnownTypeAdapterUtils.java
```diff
@@ -37,57 +42,90 @@
...
-        KNOWN_TYPE_ADAPTERS.put(short.class.getName(), "com.vimeo.stag.KnownTypeAdapters.SHORT");
-        KNOWN_TYPE_ADAPTERS.put(Integer.class.getName(), "com.vimeo.stag.KnownTypeAdapters.INTEGER");
-        KNOWN_TYPE_ADAPTERS.put(int.class.getName(), "com.vimeo.stag.KnownTypeAdapters.INTEGER");
-        KNOWN_TYPE_ADAPTERS.put(Long.class.getName(), "com.vimeo.stag.KnownTypeAdapters.LONG");
-        KNOWN_TYPE_ADAPTERS.put(long.class.getName(), "com.vimeo.stag.KnownTypeAdapters.LONG");
-        KNOWN_TYPE_ADAPTERS.put(Float.class.getName(), "com.vimeo.stag.KnownTypeAdapters.FLOAT");
-        KNOWN_TYPE_ADAPTERS.put(float.class.getName(), "com.vimeo.stag.KnownTypeAdapters.FLOAT");
-        KNOWN_TYPE_ADAPTERS.put(Double.class.getName(), "com.vimeo.stag.KnownTypeAdapters.DOUBLE");
-        KNOWN_TYPE_ADAPTERS.put(double.class.getName(), "com.vimeo.stag.KnownTypeAdapters.DOUBLE");
-        KNOWN_TYPE_ADAPTERS.put(Number.class.getName(), "com.google.gson.internal.bind.TypeAdapters.NUMBER");
```
anthonycr: I noticed that the `Number` class was added twice here (also added down below), so I removed one of them

PR #50, opened 2017-02-03
Adapter generation tests

anthonycr:
#### Summary
We can use unit testing to verify that type adapters are properly generated for every use case that Stag supports.

#### Implementation Summary
- Created unit tests for all `model` and `model1` classes in the `sample` module
- Created unit tests for all `sample_model` classes in the `sample-model` module
- Fixed a bug recently introduced (and exposed by the unit tests) where classes nested inside unsupported annotated classes were not having type adapters generated for them.
- Discovered a bug where the `ExternalModelExample2` type adapter is generated but is unable to be instantiated without generating a stack overflow.

#### How to Test
Run the unit tests in `sample` and in `sample-model`.

Note: I have not attempted to fix the stack overflow caused by the `ExternalModelExample2` type adapter.


stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/ElementUtils.java
```diff
@@ -75,7 +75,15 @@ public static boolean isEnum(@Nullable Element element) {
         return element != null && element.getKind() == ElementKind.ENUM;
     }
 
-    public static boolean isClass(@Nullable Element element){
+    public static boolean isClass(@Nullable Element element) {
         return element != null && element.getKind() == ElementKind.CLASS;
     }
+
+    public static boolean isSupportedElementKind(@Nullable Element element) {
```
yasirmhd: Good catch ! This looks good

stag-library-compiler/src/main/java/com/vimeo/stag/processor/StagProcessor.java
```diff
@@ -174,4 +172,29 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
...
+     *
+     * @param useStagElement the element to add to the
+     *                       supported type model.
+     */
+    private static void processSupportedElements(@NotNull Element useStagElement) {
+        if (ElementUtils.isSupportedElementKind(useStagElement)) {
+            TypeMirror rootType = useStagElement.asType();
+            DebugLog.log("Annotated type: " + rootType + "\n");
+            SupportedTypesModel.getInstance().getSupportedType(rootType);
+            // TODO: Should we switch to adding all elements here or continue to let AnnotatedClass.initNestedClasses do its thing 2/2/17 [AR]
```
yasirmhd: We will need to add this check in the AnnotatedClass.initNestedClasses to use ElementUtils. isSupportedElementKind

anthonycr: Good point, I'll do that - me forgetting that even while writing that comment is probably good evidence that the logic should be combined

Issue #51, opened 2017-02-03
ExternalModelExample2 Type Adapter is broken


anthonycr:
#### Issue Summary
Instantiating the Type Adapter generated for the `ExternalModelExample2` class causes a stack overflow. The cause is that the model class is generic and references itself. Note that other self referential classes such as `RecursiveClass` are generated and instantiated without error.

#### Reproduction Steps
Instantiate the type adapter with the following code and observe the stack overflow:
```java
ExternalModelExample2$TypeAdapter typeAdapter = new ExternalModelExample2$TypeAdapter(gson, factory, typeAdapter1);
```

For simplicity, uncomment the test in `ExternalModelExample2Test` and observe the problem.

#### Expected Behavior
It should properly instantiate the type adapter.

#### Actual Behavior
`StackOverflowException` occurs.

anthonycr:
Fixed by #52

PR #52, opened 2017-02-04
Self-referential generic type adapters causing stack overflows

anthonycr:
#### Ticket
https://github.com/vimeo/stag-java/issues/51 - TypeAdapters of generic classes referencing themselves caused stack overflows

#### Ticket Summary
If a generic class referenced itself (see `sample/ExternalModelExample2`), the type adapter would try to create an instance of itself in the constructor, resulting in a stack overflow.

#### Implementation Summary
The solution was to recognize that the adapter being instantiated was of the same type with the same arguments as the current class, and to instead point at `this`. This fixed the bug. I also uncommented the test case that checks this scenario.

#### How to Test
Create a generic model that references itself. Instantiate the TypeAdapter created. Observe that it does not cause a stack overflow.

### DIFF
--- a/stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
+++ b/stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
@@ -119,13 +119,12 @@ public class TypeAdapterGenerator extends AdapterGenerator {
      * This is used to generate the type token code for the types that are known.
      */
     @Nullable
-    private static String getTypeAdapterCode(@NotNull TypeMirror fieldType,
-                                             @NotNull TypeSpec.Builder adapterBuilder,
-                                             @NotNull MethodSpec.Builder constructorBuilder,
-                                             @NotNull TypeTokenConstantsGenerator typeTokenConstantsGenerator,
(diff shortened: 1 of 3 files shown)


stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
```diff
@@ -187,14 +186,19 @@ private static String getTypeAdapterCode(@NotNull TypeMirror fieldType,
...
+                        // If the field type is an external model
                         adapterCode = externalAdapterInfo.getInitializer("gson", typeAdapterCode);
                     } else {
                         ClassInfo classInfo = new ClassInfo(outerClass);
-                        int idx1 = fieldType.toString().indexOf("<");
-                        String argument = idx1 > 0 ? fieldType.toString().substring(idx1) : "";
-                        adapterCode = "new " + classInfo.getTypeAdapterQualifiedClassName() + argument +
-                                      "(gson, stagFactory" + typeAdapterCode + ")";
+                        if (classInfo.equals(mInfo)) {
+                            // In this case the adapter will be the same as the one we are generating
```
anirudhramanan: Good catch!

Issue #54, opened 2017-02-06
Duplicate Classes generated


yasirmhd:
#### Issue Summary
Duplicate classes are being generated when both the base class and derived class use @UseStag annotation and the base class exists in another module.

#### Reproduction Steps
Define a class in sample_model library

````java 
package com.vimeo.sample_model;

import com.vimeo.stag.UseStag;

@UseStag
public class BaseExternalModel {
    public String type;

    public int baseValue;

}
````

In sample app module add ExternalModelDerivedExample class
````java 
package com.vimeo.sample.model;

import com.vimeo.sample_model.BaseExternalModel;
import com.vimeo.stag.UseStag;

@UseStag
public class ExternalModelDerivedExample extends BaseExternalModel {

    public Boolean derivedBoolean;
}

````
#### Expected Behavior
It should create only one TypeAdapter for BaseExternalModel which should exist in the sample_module module only.

#### Actual Behavior
It generates two classes by BaseExternalModel$TypeAdapter in both the modules. Please check the screen shot below

What currently happens as a result of the reproduction steps?
![image](https://cloud.githubusercontent.com/assets/13655724/22642098/4f311e76-ec7f-11e6-92bb-4ddfb7d2c6e9.png)

anthonycr:
Fixed by https://github.com/vimeo/stag-java/pull/55, tests to verify added here https://github.com/vimeo/stag-java/pull/57. Closing as fixed

PR #55, opened 2017-02-06
Fixing issue #54 : Duplicate TypeAdapters being generated

yasirmhd:
By default the base classes were being added to the list of classes for which we were generating the type adapters. With this change now the list of classes for which we generate the TypeAdapters are added only by the StagProcessor. The inherited classes are kept in a separate map to reuse the AnnotatedClass object.

#### Ticket
Issues #54 

#### Ticket Summary
Duplicate TypeAdapters were being generated when the base class existed in another module and the derived class in another one and both of them were using Stag for TypeAdapter generation.

#### Implementation Summary
For the inherited types changed the implementation to cache the AnnotatedClass in another map. If that class is found to be using Stag by the StagProcessor it will be reused as well.

#### How to Test
BaseExternalModel would be regenerated if we revert the changes

### DIFF
--- a/stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
+++ b/stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
@@ -759,6 +759,9 @@ public class TypeAdapterGenerator extends AdapterGenerator {
         }
 
         AnnotatedClass annotatedClass = SupportedTypesModel.getInstance().getSupportedType(typeMirror);
+        if(null == annotatedClass) {
+            throw new IllegalStateException("The AnnotatedClass class can't be null in TypeAdapterGenerator : " + typeMirror.toString());
+        }
         Map<Element, TypeMirror> memberVariables = annotatedClass.getMemberVariables();
 
         AdapterFieldInfo adapterFieldInfo =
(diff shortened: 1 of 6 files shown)


stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
```diff
@@ -759,6 +759,9 @@ public TypeSpec getTypeAdapterSpec(@NotNull TypeTokenConstantsGenerator typeToke
         }
 
         AnnotatedClass annotatedClass = SupportedTypesModel.getInstance().getSupportedType(typeMirror);
+        if(null == annotatedClass) {
+            throw new IllegalStateException("The AnnotatedClass class can't be null in TypeAdapterGenerator : " + typeMirror.toString());
+        }
```
anthonycr: Maybe it would be better to annotate the `getSupportedType` method as `@NonNull` and throw this exception if it tries to return null in that method. What do you think?

yasirmhd: The function getSupportedType can be called from anywhere and it can throw null based on the state whether all the classes have been added or not. Also it is an independent class which does not really have the information in which order it can be called.

The reason why I added an IllegalStateException here is because the SupportedTypesModel should have been populated with this class type before calling this function.

anthonycr: Alright, that's a good point 👍

PR #57, opened 2017-02-07
Unit tests for duplicate type adapter scenario

anthonycr:
#### Summary
Added unit tests for the duplicate type adapters scenario
- a test to verify that a type adapter for `BaseExternalModel` was generated in the `sample-model` module
- a test to verify that a type adapter for the `ExternalModelDerivedExample` was generated in the `sample` module
- a test to verify that no type adapter for the `BaseExternalModel` was generated in the `sample` module

Also added some `@NotNull` annotations and added a null check to one use case of the `SupportedTypesModel`.

#### How to Test
Run the unit tests, ensure that they all pass (The build will do this).


anthonycr:
thanks for the bug fix 👍 

I'll remember to mention unit tests the next time there are new models added

yasirmhd:
👍 

@anthonycr Can you please publish a SNAPSHOT version so that we can continue to validate this on a regular basis

anthonycr:
@yasirmhd I'll work on getting a snapshot out

PR #59, opened 2017-02-08
Cleaning up GenericClassInfo class and cleaning up utils classes

anthonycr:
#### Summary
- Removed unnecessary field from GenericClassInfo.
- Fixed NPE lint warnings in TypeUtils and ElementUtils.
- Cleaned up TypeUtils and ElementUtils APIs.
- Deleted unnecessary unit tests.

### DIFF
--- a/stag-library-compiler/src/main/java/com/vimeo/stag/processor/StagProcessor.java
+++ b/stag-library-compiler/src/main/java/com/vimeo/stag/processor/StagProcessor.java
@@ -115,9 +115,10 @@ public final class StagProcessor extends AbstractProcessor {
 
         String packageName = getOptionalPackageName(processingEnv);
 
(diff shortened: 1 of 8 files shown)


yasirmhd:
LGTM. The reason for GenericClassInfo having that extra parameter was for some future work, but this is fine as we can always bring it back when we implement it.

anthonycr:
Alright, I initially was doing this work because I didn't like that other field (`mNumArguments`) was hardcoded due to its error proneness. I found a way to generate it dynamically using a `TypeElement` object, `typeElement.getTypeParameters().size()` and I originally wrote the code to replace the hardcoded field. In the end, I decided not to add that code in, and instead just remove it because the field was unused. 

When we need the field in the future, now we know we can get it dynamically.

PR #61, opened 2017-02-15
Improved Performance

yasirmhd:
We were doing JsonReader.peek which is an expensive function. Reducing this improves the performance significantly.

### DIFF
--- a/stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
+++ b/stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
@@ -34,6 +34,7 @@ import com.squareup.javapoet.ParameterizedTypeName;
 import com.squareup.javapoet.TypeName;
 import com.squareup.javapoet.TypeSpec;
 import com.squareup.javapoet.TypeVariableName;
+import com.vimeo.stag.KnownTypeAdapters;
 import com.vimeo.stag.KnownTypeAdapters.ArrayTypeAdapter;
 import com.vimeo.stag.KnownTypeAdapters.ListTypeAdapter;
 import com.vimeo.stag.KnownTypeAdapters.MapTypeAdapter;
@@ -268,11 +269,8 @@ public class TypeAdapterGenerator extends AdapterGenerator {
             boolean isPrimitive = TypeUtils.isSupportedPrimitive(variableType);
 
             if(isPrimitive) {
-                builder.addCode("\t\t\t\t" + TypeUtils.getObjectForPrimitive(variableType) + " " + variableName + " = " +
-                        adapterFieldInfo.getAdapterAccessor(elementValue) + ".read(reader);\n");
-                builder.addCode("\t\t\t\tif(null != " + variableName + ") {\n");
-                builder.addCode("\t\t\t\t\tobject." + variableName + " = " + variableName + ";\n");
-                builder.addCode("\t\t\t\t}");
+                builder.addCode("\t\t\t\tobject." + variableName + " = " +
+                        adapterFieldInfo.getAdapterAccessor(elementValue) + ".read(reader, object." + variableName +  ");");
 
             } else {
                 builder.addCode("\t\t\t\tobject." + variableName + " = " +
@@ -616,6 +614,8 @@ public class TypeAdapterGenerator extends AdapterGenerator {
             if (hasUnknownGenericField && TypeUtils.containsTypeVarParams(fieldType)) {
                 adapterAccessor = getAdapterForUnknownType(fieldType, adapterBuilder, constructorBuilder,
                                                            typeTokenConstantsGenerator, typeVarsMap, result);
+            } else if(KnownTypeAdapterUtils.hasNativePrimitiveTypeAdapter(fieldType)) {
+                adapterAccessor = KnownTypeAdapterUtils.getNativePrimitiveTypeAdapter(fieldType);
             } else {
                 adapterAccessor = getAdapterAccessor(fieldType, adapterBuilder, constructorBuilder,
                                                      typeTokenConstantsGenerator, typeVarsMap, stagGenerator,
--- a/stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
+++ b/stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
@@ -264,8 +264,21 @@ public class TypeAdapterGenerator extends AdapterGenerator {
                 }
             }
 
-            builder.addCode("\t\t\t\tobject." + variableName + " = " +
-                            adapterFieldInfo.getAdapterAccessor(elementValue) + ".read(reader);");
+            String variableType = element.getValue().toString();
+            boolean isPrimitive = TypeUtils.isSupportedPrimitive(variableType);
+
+            if(isPrimitive) {
+                builder.addCode("\t\t\t\t" + TypeUtils.getObjectForPrimitive(variableType) + " " + variableName + " = " +
+                        adapterFieldInfo.getAdapterAccessor(elementValue) + ".read(reader);\n");
+                builder.addCode("\t\t\t\tif(null != " + variableName + ") {\n");
+                builder.addCode("\t\t\t\t\tobject." + variableName + " = " + variableName + ";\n");
+                builder.addCode("\t\t\t\t}");
+
+            } else {
+                builder.addCode("\t\t\t\tobject." + variableName + " = " +
+                        adapterFieldInfo.getAdapterAccessor(elementValue) + ".read(reader);");
+            }
+
 
             builder.addCode("\n\t\t\t\tbreak;\n");
             runIfAnnotationSupported(element.getKey().getAnnotationMirrors(), new Runnable() {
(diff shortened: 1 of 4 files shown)


yasirmhd:
@anthonycr : With this PR + the changes I have raised with Gson (https://github.com/google/gson/pull/1015) STAG is doing better than LoganSquare based on their [BenchmarkDemo](https://github.com/bluelinelabs/LoganSquare) 


image

anthonycr:
Awesome! Gonna take a look at this soon

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
```diff
@@ -245,11 +246,6 @@ private static MethodSpec getReadMethodSpec(@NotNull TypeName typeName,
                         "();\n" +
                         "\twhile (reader.hasNext()) {\n" +
                         "\t\tString name = reader.nextName();\n" +
-                        "\t\tcom.google.gson.stream.JsonToken jsonToken = reader.peek();\n" +
-                        "\t\tif (jsonToken == com.google.gson.stream.JsonToken.NULL) {\n" +
-                        "\t\t\treader.skipValue();\n" +
-                        "\t\t\tcontinue;\n" +
-                        "\t\t}\n" +
```
anthonycr: I'm interested in what the rational is for this besides performance improvement. I see it as a potential problem because we have two potential flows for primitives (objects would be fine because we would resort to the object type adapters which check for null).
1. Parsed via a type adapter that is in the gson `TypeAdapters` class: e.g. booleans are parsed using `TypeAdapters.BOOLEAN`. This adapter does a null check and returns null of the token is null. We then do a null check and optionally assign the field depending on if the item returned was null. This means the field remains uninitialized.
2. Parsed via a type adapter that is in `KnownTypeAdapters`. This adapter does not do a null check and will throw an illegal state exception because the peeked type did not match the type that it wanted (since the fields are ultimately assigned via `reader.nextType`.

Am I misunderstanding something? I think we should be consistent for all primitives

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
```diff
@@ -268,8 +264,21 @@ private static MethodSpec getReadMethodSpec(@NotNull TypeName typeName,
...
-                            adapterFieldInfo.getAdapterAccessor(elementValue) + ".read(reader);");
+            String variableType = element.getValue().toString();
+            boolean isPrimitive = TypeUtils.isSupportedPrimitive(variableType);
+
+            if(isPrimitive) {
+                builder.addCode("\t\t\t\t" + TypeUtils.getObjectForPrimitive(variableType) + " " + variableName + " = " +
+                        adapterFieldInfo.getAdapterAccessor(elementValue) + ".read(reader);\n");
+                builder.addCode("\t\t\t\tif(null != " + variableName + ") {\n");
+                builder.addCode("\t\t\t\t\tobject." + variableName + " = " + variableName + ";\n");
+                builder.addCode("\t\t\t\t}");
```
anthonycr: I see this prevents us from getting an exception if the gson primitive type adapter returns null, that's great. See above for my comment about consistency between gson primitive type adapters and our type adapters

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
```diff
@@ -115,7 +115,7 @@ public void write(JsonWriter out, Short value) throws IOException {
         @Override
         public Integer read(JsonReader in) throws IOException {
             try {
-                return in.nextInt();
+                return Integer.valueOf(in.nextInt());
```
yasirmhd: The warning that Java is giving is to use the primitive type directly instead of casting it into an object  - PFA the details around the warning

image


Now looking at the comments and implementation for Integer.valueOf

image

Integer, Long, etc are immutable types. So for the purpose of optimisation the platform caches some of these instances so that an allocation need not be done every time. For instance in this case they are caching values between -128 and 127.

Given the number of times this was being called profiling told us that we should use Integer.valueOf instead of the constructor - which is also the recommended way of doing it

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
```diff
@@ -77,7 +77,7 @@ private KnownTypeAdapters() {
         public Byte read(JsonReader in) throws IOException {
             try {
                 int intValue = in.nextInt();
-                return (byte) intValue;
+                return Byte.valueOf((byte) intValue);
```
yasirmhd: Same as the example in Integer. Kindly let me know if it is OK

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
yasirmhd: Primary reason is performance


1 - uninitialized fields should not be a problem as that can happen if the server does not send the data. Previous implementation also would have created an object with similar state.
2. Not if we ensure all primitive type adapters in KnownTypeAdapters are null safe. 

But with this you have given me an idea. I will implement methods in KnownTypeAdapters of the sort :

int read(JsonReader in, int default value)

Which will be called as :

obj.field = KnownTypeAdapters.read(reader, obj.field)

In the event of the response being null we will either set it to a default value or its existing value

yasirmhd: Let me know what you think and I will make the required changes

yasirmhd:
@anthonycr thanks for your thorough review. I have updated the code so that we ensure the nullity checks are in place. We will always use our custom classes for primitive type parsing. This will also avoid the unnecessary boxing and object creations. 

For the assignment part, if the value is null it will just take whatever the default value is - which again is the default gson behaviour as well.

Kindly review this

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
anthonycr: Okay, yes, the reason for using `Integer.valueOf` makes sense because caching the objects. However, the whole point of the warning is because the compiler automatically applies the caching post java 1.5. Calling `Integer integer = Integer.valueOf(1)` will produce the same java bytecode as `Integer integer = 1`, and for this reason the warning exists.

I've written a unit test demonstrating that this is the case. Here we test object reference equality rather than using the `equals` method to ensure that all objects are pointing to the same memory location. If you run this test, you should see that when in both cases, the integer is <= 127, the object is cached by the `Integer` class. When it's greater, it allocates a new object. I've tested this using JUnit, on the Emulator, and on my Nexus 5X and verified the behavior is the same everywhere.

```java
public class AutoBoxingTest {

    @Test
    public void testBoxing_equal(){
        Object integer1 = Integer.valueOf(1);
        Object integer2 = Integer.valueOf(1);

        Assert.assertTrue(integer1 == integer2);
    }

    @Test
    public void testBoxing_notEqual(){
        Object integer1 = Integer.valueOf(Integer.MAX_VALUE);
        Object integer2 = Integer.valueOf(Integer.MAX_VALUE);

        Assert.assertFalse(integer1 == integer2);
    }

    @Test
    public void testNoBoxing_notEqual(){
        Object integer1 = Integer.MAX_VALUE;
        Object integer2 = Integer.MAX_VALUE;

        Assert.assertFalse(integer1 == integer2);
    }

    @Test
    public void testNoBoxing_equal(){
        Object integer1 = 127;
        Object integer2 = 127;

        Assert.assertTrue(integer1 == integer2);
    }

}
```

For this reason I am proposing the removal of the autoboxing, as it makes the code unnecessarily complicated.

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
anthonycr: 1. True, the option of the server not sending data slipped my mind. And if you truly want to enforce a field always being filled, you can just put the `@NonNull` annotation on your field and stag will verify it. Okay, so that's not a problem.
2. Okay that's a good idea then, as we will be comparable with gson. 

The static method with a default value is also a good idea 👍

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
yasirmhd: Thanks for the details info here. I have made the changes.

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
```diff
@@ -176,6 +294,72 @@ public void write(JsonWriter out, Double value) throws IOException {
...
+    /**
+     * Helper class for {@link boolean}.
+     */
+    public static final class PrimitiveBooleanTypeAdapter{
+
+        public static boolean read(JsonReader in, boolean defaultValue) throws IOException {
+            if (in.peek() == JsonToken.NULL) {
+                in.nextNull();
+                return defaultValue;
+            } else if (in.peek() == JsonToken.STRING) {
```
yasirmhd: Will update in a few minutes

yasirmhd: done. Thanks !

yasirmhd:
Agreed on Object.valueOf. 

image


This one remains because if I would have used the char directly it would have gone to the method which takes long as the parameter.

Issue #63, opened 2017-02-17
PrimitiveCharArrayAdapter is broken


anthonycr:
#### Issue Summary
The char array type adapter is broken. It writes an array of numbers to json since it writes the characters as numbers, but tries to read the array out of the json as a string. This is broken. See `KnownTypeAdaptersTest.testForPrimitiveArrayCharacterTypeAdapter()` for a test case that breaks the adapter. 

#### Reproduction Steps
```java
char[] value = new char[5];

value[0] = 'a';
value[1] = 'b';
value[2] = 'c';
value[3] = 'd';
value[4] = 'e';

StringWriter stringWriter = new StringWriter();
KnownTypeAdapters.PrimitiveCharArrayAdapter.write(new JsonWriter(stringWriter), value);
String jsonString = stringWriter.toString();

char[] readValue = KnownTypeAdapters.PrimitiveCharArrayAdapter.read(new JsonReader(new StringReader(jsonString)));

// This is broken
Assert.assertArrayEquals(value, readValue);
```

#### Expected Behavior
The char array should be read from json correctly.

#### Actual Behavior
An exception is thrown:

```
java.lang.IllegalStateException: Expected a string but was BEGIN_ARRAY at line 1 column 2 path $
	at com.google.gson.stream.JsonReader.nextString(JsonReader.java:825)
	at com.google.gson.internal.bind.TypeAdapters$16.read(TypeAdapters.java:418)
	at com.google.gson.internal.bind.TypeAdapters$16.read(TypeAdapters.java:406)
	at com.google.gson.TypeAdapter$1.read(TypeAdapter.java:199)
	at com.vimeo.stag.KnownTypeAdapters$PrimitiveCharArrayAdapter.read(KnownTypeAdapters.java:688)
```

anirudhramanan:
Will be fixed in PR #65

PR #65, opened 2017-02-18
Char Type Adapter Fix

anirudhramanan:
Refer #63.


stag-library/src/test/java/com/vimeo/stag/KnownTypeAdaptersTest.java
```diff
@@ -204,4 +204,39 @@ public void testForJsonArrayTypeAdapter() throws Exception {
             Assert.assertEquals(jsonArray.get(i), readValue.get(i));
         }
     }
+
+    @Test
+    public void testForPrimitiveArrayCharacterTypeAdapter() throws Exception {
```
anirudhramanan: same test case that has been added in PR #64

anirudhramanan: same test case that has been added in PR #64

PR #67, opened 2017-02-19
Enhancements ( Boolean Parsing and more detailed exception message )

anirudhramanan:


stag-library/src/test/java/com/vimeo/stag/KnownTypeAdaptersTest.java
```diff
@@ -450,8 +452,36 @@ public void testForPrimitiveBooleanTypeAdapter() throws Exception {
...
+
+        booleanString = "false";
+        // call the TypeAdapter#read method
+        readValue = KnownTypeAdapters.PrimitiveBooleanTypeAdapter.read(new JsonReader(new StringReader(booleanString)), false);
+        Assert.assertEquals(false, readValue);
+
+        booleanString = "abcde"; // not valid, throw exception
+        // call the TypeAdapter#read method
+        readValue = KnownTypeAdapters.PrimitiveBooleanTypeAdapter.read(new JsonReader(new StringReader(booleanString)), false);
+        Assert.assertEquals(false, readValue);
```
anthonycr: This test throws an exception and the rest of the cases don't run as method execution is halted. You should either break this case and the `booleanString = "2";` case out into their own test methods that are annotated with `@Test (expected = MalformedJsonException.class)`, or you need to write the failing tests as follows:

```java
boolean exceptionWasThrown = false;
try {
booleanString = "abcde"; // not valid, throw exception
// call the TypeAdapter#read method
readValue = KnownTypeAdapters.PrimitiveBooleanTypeAdapter.read(new JsonReader(new StringReader(booleanString)), false);
} catch (MalformedJsonException exception) {
    exceptionWasThrown = true;
}
Assert.assertTrue(exceptionWasThrown);
```

stag-library/src/test/java/com/vimeo/stag/KnownTypeAdaptersTest.java
```diff
@@ -450,8 +452,36 @@ public void testForPrimitiveBooleanTypeAdapter() throws Exception {
 
         // call the TypeAdapter#read method
         boolean readValue = KnownTypeAdapters.PrimitiveBooleanTypeAdapter.read(new JsonReader(new StringReader(jsonString)), false);
-
         Assert.assertEquals(value, readValue);
+
+        String booleanString = "true";
+        // call the TypeAdapter#read method
+        readValue = KnownTypeAdapters.PrimitiveBooleanTypeAdapter.read(new JsonReader(new StringReader(booleanString)), false);
+        Assert.assertEquals(true, readValue);
```
anthonycr: We should use `Assert.assertTrue`

stag-library/src/test/java/com/vimeo/stag/KnownTypeAdaptersTest.java
```diff
@@ -450,8 +452,36 @@ public void testForPrimitiveBooleanTypeAdapter() throws Exception {
...
+
+        booleanString = "0"; // 0 is treated as false
+        // call the TypeAdapter#read method
+        readValue = KnownTypeAdapters.PrimitiveBooleanTypeAdapter.read(new JsonReader(new StringReader(booleanString)), false);
+        Assert.assertEquals(false, readValue);
+
+        booleanString = "2"; // not valid, throw exception
+        // call the TypeAdapter#read method
+        readValue = KnownTypeAdapters.PrimitiveBooleanTypeAdapter.read(new JsonReader(new StringReader(booleanString)), false);
```
anthonycr: This case also needs the new try/catch logic

stag-library/src/test/java/com/vimeo/stag/KnownTypeAdaptersTest.java
```diff
@@ -450,8 +452,36 @@ public void testForPrimitiveBooleanTypeAdapter() throws Exception {
...
+
+        booleanString = "1"; // 1 is treated as true
+        // call the TypeAdapter#read method
+        readValue = KnownTypeAdapters.PrimitiveBooleanTypeAdapter.read(new JsonReader(new StringReader(booleanString)), false);
+        Assert.assertEquals(true, readValue);
+
+        booleanString = "0"; // 0 is treated as false
+        // call the TypeAdapter#read method
+        readValue = KnownTypeAdapters.PrimitiveBooleanTypeAdapter.read(new JsonReader(new StringReader(booleanString)), false);
+        Assert.assertEquals(false, readValue);
```
anthonycr: Use `Assert.assertFalse`, also we should pass in `true` as the default value for this read so that we know that the default value isn't being used

stag-library/src/test/java/com/vimeo/stag/KnownTypeAdaptersTest.java
```diff
@@ -450,8 +452,36 @@ public void testForPrimitiveBooleanTypeAdapter() throws Exception {
...
+
+        booleanString = "abcde"; // not valid, throw exception
+        // call the TypeAdapter#read method
+        readValue = KnownTypeAdapters.PrimitiveBooleanTypeAdapter.read(new JsonReader(new StringReader(booleanString)), false);
+        Assert.assertEquals(false, readValue);
+
+        booleanString = "1"; // 1 is treated as true
+        // call the TypeAdapter#read method
+        readValue = KnownTypeAdapters.PrimitiveBooleanTypeAdapter.read(new JsonReader(new StringReader(booleanString)), false);
+        Assert.assertEquals(true, readValue);
```
anthonycr: use `Assert.assertTrue` instead

stag-library/src/test/java/com/vimeo/stag/KnownTypeAdaptersTest.java
```diff
@@ -450,8 +452,36 @@ public void testForPrimitiveBooleanTypeAdapter() throws Exception {
...
+
+        String booleanString = "true";
+        // call the TypeAdapter#read method
+        readValue = KnownTypeAdapters.PrimitiveBooleanTypeAdapter.read(new JsonReader(new StringReader(booleanString)), false);
+        Assert.assertEquals(true, readValue);
+
+        booleanString = "false";
+        // call the TypeAdapter#read method
+        readValue = KnownTypeAdapters.PrimitiveBooleanTypeAdapter.read(new JsonReader(new StringReader(booleanString)), false);
+        Assert.assertEquals(false, readValue);
```
anthonycr: Use `Assert.assertFalse`, also we should pass in `true` as the default value for this read so that we know that the default value isn't being used

stag-library/src/main/java/com/vimeo/stag/KnownTypeAdapters.java
```diff
@@ -336,27 +338,36 @@ public static void write(JsonWriter out, char value) throws IOException {
...
         public static boolean read(JsonReader in, boolean defaultValue) throws IOException {
             JsonToken peek = in.peek();
             if (peek == JsonToken.NULL) {
                 in.nextNull();
                 return defaultValue;
-            } else if (peek == JsonToken.STRING) {
+            } else if (peek == JsonToken.STRING || peek == JsonToken.NUMBER) {
                 // support strings for compatibility with GSON 1.7
-                return Boolean.parseBoolean(in.nextString());
+                return Boolean.parseBoolean(booleanAsString(in.nextString()));
```
anthonycr: As mentioned below, we can get rid of the `Boolean.parseBoolean` call if `booleanAsString` returned a boolean. In that case it should probably be renamed to `stringToBoolean`

yasirmhd: Should we not just keep this simple and have it as 

((name != null) && (name.equalsIgnoreCase("true") || name.equals("1")))


image

anthonycr: I agree that simplifying it even further has benefit. However, I also see benefit for checking if the `VALID_BOOLEAN_AS_STRING` list contains the string, because if it doesn't I think it's appropriate to throw an exception rather than just default to false.

yasirmhd: I have concerns with throwing exception here. I can't corelate this to a use case that we have but it might break things for some someone where they were returning true along with other values which do not exist in the set of valid boolean values. Anyone doing this would not be doing the right thing but we will deviate from the Gson behavior here.

anthonycr:
I'm gonna close this as it's no longer needed. We can reference it if we decide to change our minds.

PR #74, opened 2017-03-03
Final fields should be raised as compile-time errors

mcumings:
#### Ticket Summary

Previous behavior was to silently omit the final fields from the set of
fields supported by the generated type adapter.

Also adds functional testing capabilities primarily to test compilation
failure scenarios.

#### Implementation Summary

Implementation is primarily in the creation of compilation failure report testing.

#### How to Test

Attempt to use final or private fields in a model class and observe the beautiful compile-time
error. 😄   Note the transition from throwing exceptions upon failure to using the actual annotation processor's Messager instance which points the user at the location in the code (i.e., linked in the IDE etc.).

### DIFF
--- a/stag-library-compiler/src/test/java/com/vimeo/stag/processor/AbstractAnnotationProcessorTest.java
+++ b/stag-library-compiler/src/test/java/com/vimeo/stag/processor/AbstractAnnotationProcessorTest.java
@@ -1,0 +1,155 @@
+package com.vimeo.stag.processor;
+
+/*
+ * @(#)AbstractAnnotationProcessorTest.java     5 Jun 2009
+ *
+ * Copyright © 2010 Andrew Phillips.
+ *
+ * ====================================================================
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ====================================================================
+ */
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Locale;
+
+import javax.annotation.processing.Processor;
+import javax.tools.Diagnostic;
+import javax.tools.Diagnostic.Kind;
+import javax.tools.DiagnosticCollector;
+import javax.tools.JavaCompiler;
+import javax.tools.JavaCompiler.CompilationTask;
+import javax.tools.JavaFileObject;
+import javax.tools.StandardJavaFileManager;
+import javax.tools.ToolProvider;
+
+import static com.google.common.base.Throwables.propagate;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * A base test class for {@link Processor annotation processor} testing that
+ * attempts to compile source test cases that can be found on the classpath.
+ *
+ * @author aphillips
+ * @since 5 Jun 2009
+ */
+abstract class AbstractAnnotationProcessorTest {
+    private static final String SOURCE_FILE_SUFFIX = ".java";
+    private static final JavaCompiler COMPILER = ToolProvider.getSystemJavaCompiler();
+
+    /** @return the processor instances that should be tested */
+    protected abstract Collection<Processor> getProcessors();
+
+    protected List<Diagnostic<? extends JavaFileObject>> compileTestCase(String ... classNames) {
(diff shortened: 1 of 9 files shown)


yasirmhd:
+1 This would be very helfpul

yasirmhd:
On second thoughts if we do print the right errors ,  I guess the current behavior might be more appropriate. @anthoncr what are your thoughts on  this ?

mcumings:
@yasirmhd - I'm not 100% sure what you are suggesting.  I think you are suggesting that we should use the `Messager` to raise the final fields at the warning level instead of at the error level and therefore allow compilation to succeed with the missing fields?

yasirmhd:
@mcumings  I am suggesting that we do keep the code of printing the error using the Messager in these case and throw a RunTimeException.

This way the build will fail with an exception.

mcumings:
Okay, thanks for the clarification.

If the `Messager` emits an error the build will fail.  The question then becomes whether or not it is appropriate to have an exception raised as well.  In scenarios that the annotation processor encounters which amount to a validation of the usage scenario then the `Messager` approach without an associated exception is the most appropriate.  Exceptions should only be raised by an annotation processor if there is an error in the logic of the processor (e.g., NPE, Preconditions, etc.).

yasirmhd:
Sure. Thanks for the explanation. I guess we should be good with this one then

stag-library-compiler/src/test/java/com/vimeo/stag/processor/StagProcessorFunctionalTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.Collection;
+import java.util.Collections;
+
+import javax.annotation.processing.Processor;
+import javax.tools.Diagnostic;
+
+public class StagProcessorFunctionalTest extends AbstractAnnotationProcessorTest {
+
+    private StagProcessor processor;
```
anthonycr: This appears unused outside of the `getProcessors` method and can be turned into a local variable.

stag-library-compiler/src/test/java/com/vimeo/stag/processor/CleanableJavaFileManager.java
```diff
@@ -0,0 +1,103 @@
...
+
+import javax.tools.FileObject;
+import javax.tools.JavaFileManager;
+import javax.tools.JavaFileObject;
+
+/**
+ * {@link JavaFileManager} implementation which tracks created files and allows them to be
+ * systematically deleted.
+ */
+class CleanableJavaFileManager implements JavaFileManager {
```
anthonycr: Is this file also from the same source as `AbstractAnnotationProcessorTest.java`? If so, we should probably have the same license header as the source

stag-library-compiler/src/test/java/com/vimeo/stag/processor/StagProcessorFunctionalTest.java
```diff
@@ -0,0 +1,39 @@
...
+        return Collections.<Processor>singletonList(processor);
+    }
+
+    /**
+     * Ensure that final fields result in compile-time errors to prevent silent omission of fields
+     * from generated type adapters.
+     */
+    @Test
+    public void finalFieldsInAnnotatedClassReportsAsAnError() throws Exception {
+        assertCompilationReturned(Diagnostic.Kind.ERROR, 8, compileTestCase("bad/FinalFields"));
```
anthonycr: Very clean and tests work well 👍

stag-library-compiler/src/test/java/com/vimeo/stag/processor/AbstractAnnotationProcessorTest.java
```diff
@@ -0,0 +1,155 @@
...
+import static com.google.common.base.Throwables.propagate;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * A base test class for {@link Processor annotation processor} testing that
+ * attempts to compile source test cases that can be found on the classpath.
+ *
+ * @author aphillips
+ * @since 5 Jun 2009
```
anthonycr: is there a URL we could link to in this file, or should we just leave it as is since attribution seems pretty clear?

stag-library-compiler/src/test/java/com/vimeo/stag/processor/CleanableJavaFileManager.java
mcumings: It is not.  I spun this up in response to test artifacts (i.e., generated source) being left in the module's root dir.  It actually still leaves empty directories around but git doesn't care since it doesn't do directories natively.

anthonycr: Alright cool 👍

stag-library-compiler/src/test/java/com/vimeo/stag/processor/AbstractAnnotationProcessorTest.java
mcumings: I think it's a good idea.  👍

stag-library-compiler/src/test/java/com/vimeo/stag/processor/CleanableJavaFileManager.java
mcumings: I went ahead and added the Vimeo/MIT license header to this and the other new file to make that more explicitly clear.

anthonycr:
I think we are good to merge

Issue #73, opened 2017-03-03
Stag generates type adapters for non-annotated nested classes


mcumings:
#### Issue Summary

Given the class:
```
@UseStag
public class NestedClassWithoutUseStag {

    String field;

    /**
     * No adapter should be created for this class.
     */
    public static class NestedWithoutUseStag {

    }
}
```

Stag is generating TypeAdapter implementations for the outer class (`NestedClassWithoutUseStag`) as expected but is also generating a TypeAdapter for the nested, unannotated class (`NestedWithoutUseStag`).  This is unexpected as it makes a fairly broad assumption of the nature of the nested class in that it assumes it is structurally related to the model/entity class and - for instance - is not a nested JsonDeserializer / JsonSerializer / TypeAdapter implementation.

This additionally causes problems if some of the fields of the top level class are of types defined as nested inner classes but have a custom type adapter registered with Gson.  Because Stag is automatically generating the type adapters for the nested class the nested class becomes known to Stag and the Gson type adapter lookup is bypassed.

#### Expected Behavior

Expected behavior would be that classes need to be explicitly annotated with @UseStag in order for a TypeAdapter to be generated.

#### Actual Behavior

Nested classes have TypeAdapters generated as with the FieldOption.ALL option.  This looks to be explicitly added in  88ab69f81aa15ae4f8ad321ba411a8c909e49c4e with commit comment `Fixed backwards incompatibility, fixed nested classes always defaulting to fieldoption.all`.

Raising this as an issue so that we can discuss whether this makes sense and/or address the issue via other means.

mcumings:
@anthonycr - Want to get your opinion on this as you have some context on the original spec and the addition of this backwards compatibility change.

One potential solution might be to add a `FieldOptions.INHIBIT` to specifically mark classes which should under no circumstances cause code to be generated but this feels clunky to me.  It would make better sense to me to drop the backwards compatibility change, given that @UseStag was intended to be allowed as a breaking change.

anthonycr:
@mcumings yeah, so the original spec was that inner classes would inherit the annotation from the outer classes. I'm not sure it was ever discussion about it. I think what happened is that with the original `@UseStag` implementation the inner classes were also picked up by the annotation processor and processed, probably by accident. Really it should have been viewed as a bug, but after it was already in, I thought it was convenient and just decided to fix the bug seen in that commit. 

[...]

PR #76, opened 2017-03-06
Unannotated nested classes should not generate type adapters.

mcumings:
#### Ticket Summary

Addresses https://github.com/vimeo/stag-java/issues/73 by ensuring that adapters are only generated for classes annotated with `@UseStag`.

#### Implementation Summary

Changes test expectations to match the new agreed upon direction, effectively filtering
the supported `ElementKind.CLASS` and `ElementKind.ENUM` types by those which have explicitly added the `@UseStag`
annotation.

I vacillated a bit on whether inherited annotations should cause type adapter generation
and settled on this approach as it seemed reasonable to me.

#### How to Test

Run the unit tests. ;)


stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/model/AnnotatedClass.java
```diff
@@ -66,7 +66,7 @@
 
         FieldOption fieldOption = useStag != null ? useStag.value() : null;
         if (fieldOption == null) {
-            useStag = findParentUseStagAnnotation(element);
+            useStag = ElementUtils.findAnnotation(UseStag.class, element);
```
mcumings: Note: This represents a subtle functional change where we consult the type hierarchy rather than the definition location hierarchy to determine the field options.

stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/TypeUtils.java
```diff
@@ -547,7 +547,30 @@ public static String getClassNameFromTypeMirror(@NotNull TypeMirror typeMirror)
...
+    /**
+     * Convert the provided {@link TypeMirror} into an {@link Element} instance.  This method
+     * call assumes that the provided {@link TypeMirror} is constrained to only known supported
+     * types.  As a result it will guarantee non-null result values.
+     *
+     * @param typeMirror type mirror to convert
+     * @return element representation of the type mirror
+     */
     @NotNull
+    public static Element getElementFromSupportedTypeMirror(@NotNull TypeMirror typeMirror) {
```
mcumings: Note: The original methods was annotated with `@NotNull` even though `getUtils().asElement(typeMirror)` can technically return `null` values.  The reason it didn't was only due to the fact that we only ever passed in supported types.  This method is an attempt to make this more explicit.

stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/ElementUtils.java
```diff
@@ -72,6 +76,42 @@ public static String getPackage(@NotNull TypeMirror type) {
...
+            return annotatedElement;
+
+        if (element instanceof TypeElement) {
+            TypeElement typeElement = (TypeElement) element;
+            TypeMirror superTypeMirror = typeElement.getSuperclass();
+            if (superTypeMirror != null) {
+                Element superTypeElement = TypeUtils.getElementFromTypeMirror(superTypeMirror);
+                return superTypeElement == null ? null : findAnnotation(annotationType, superTypeElement);
+            }
+        }
```
anthonycr: Since we aren't supporting classes that aren't annotated now (unless they have the GsonAdapterKey annotation for the time being), shouldn't this block of code be removed?

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/model/AnnotatedClass.java
anthonycr: Understood, this change is in keeping with the discussion on the related issue 👍

stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/ElementUtils.java
```diff
@@ -72,6 +76,42 @@ public static String getPackage(@NotNull TypeMirror type) {
...
+
+        return null;
+    }
+
+    public static <T extends Annotation> boolean isAnnotatedWith(
```
anthonycr: minor: would be nice to have a little javadoc on this

stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/ElementUtils.java
```diff
@@ -72,6 +76,42 @@ public static String getPackage(@NotNull TypeMirror type) {
...
+            @NotNull Class<T> annotationClass,
+            @Nullable Element element) {
+        T annotation = element == null ? null : element.getAnnotation(annotationClass);
+        return annotation != null;
```
anthonycr: minor: this whole method could be simplified to:
```java
return element != null && element.getAnnotation(annotationClass) != null;
```

stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/ElementUtils.java
```diff
@@ -83,7 +123,8 @@ public static boolean isSupportedElementKind(@Nullable Element element) {
             return false;
         }
         ElementKind elementKind = element.getKind();
-        return elementKind == ElementKind.CLASS || elementKind == ElementKind.ENUM;
+        return (elementKind == ElementKind.CLASS || elementKind == ElementKind.ENUM)
+                && isAnnotatedWith(UseStag.class, element);
```
anthonycr: 👍

sample-java-model/build.gradle
```diff
@@ -0,0 +1,21 @@
+plugins {
+    id "java"
+}
```
mcumings: Will do.  I had the new plugins form in place originally since I though I was going to need the apt plugin, but alas, compileOnly is natively supported now!  w00t

stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/ElementUtils.java
mcumings: In the code at the point of use the search is performed in order to determine the field options.  There is currently a comment there stating: `TODO: Do not default to FieldOption.SERIALIZED_NAME, instead set to null and fail after removing @GsonAdapterKey 1/30/17 [AR]`

Perhaps this goes away post-merge.  I'll push a commit for all other feedback and re-evaluate post-merge.

mcumings: Okay, I think this is still valid - at least until you merge the removal of `@GsonAdapterKey`.  The code in question is still needed to walk up the inheritance hierarchy to determine what field options are set.  Do I read that correctly?

anthonycr: Yeah I think it's fine to leave for right now. The fact that there's an open PR for the removal of gson adapter key is kind of muddying some of this logic, but we can remove it if it's necessary in that PR. Fine as is.

Issue #47, opened 2017-01-31
Allow 1's and 0's to be parsed as booleans


anthonycr:
#### Issue Summary
Currently, binary numbers are not supported by Stag for parsing into boolean fields. This is not something that Gson supports. Proposed spec:
- for field `boolean`
- numbers `1` and `0` will be properly parsed into `true` and `false`.
- other numbers will throw a `JsonSyntaxException`

#### Reproduction Steps
create a field
````
@SerializedName("boolean_field")
boolean myField;
````
parse the following data into that field
````
{
"boolean_field":"1"
}
````

#### Expected Behavior
We would expect that binary numbers would be accepted as true and false.

#### Actual Behavior
Observe a `JsonSyntaxException` being thrown and parsing halted.

anthonycr:
Refer to this Gson class: https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java

Arguably, we could just be referencing these adapters instead of creating our own in `KnownTypeAdapters`

anthonycr:
I originally thought this was supported by Gson, but it is not, so I've decided this is not a bug, but is a potential enhancement.

anthonycr:
Implemented in this PR: https://github.com/vimeo/stag-java/pull/67

I've decided this is not needed at this time, and am closing this issue.

PR #77, opened 2017-03-08
Removing GsonAdapterKey

anthonycr:
#### Summary
Removing the deprecated `@GsonAdapterKey` annotation. `@SerializedName` is not the only supported key naming annotation.

#### Implementation Summary
- Removed the annotation.
- Removed classes testing support for both `@GsonAdapterKey` and `@SerializedName`.
- Cleaned up code in the StagProcessor and refactored for clarity.
- Renamed and refactored some generator methods so that code generator classes all follow the same style: `createTypeSpec` method that returns the TypeSpec which can then be handed off to the filer to create a file.
- Added new CodeStyle.xml
- Updated gradle to latest distribution.

#### How to Test
Run the tests!

NOTE: This is a breaking change to stag and will require at least a minor version bump in the next release.


stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/AdapterGenerator.java
```diff
@@ -58,16 +57,26 @@ static String getJsonName(@NotNull Element element) {
...
+     *
+     * @param typeTokenConstantsGenerator the generator for the TypeTokenConstants class.
+     * @param stagGenerator               the generator for the Stag class.
+     * @return A non null {@link TypeSpec} describing the adapter class.
+     */
     @NotNull
-    public abstract TypeSpec getTypeAdapterSpec(
-            @NotNull TypeTokenConstantsGenerator typeTokenConstantsGenerator,
-            @NotNull StagGenerator stagGenerator);
+    public abstract TypeSpec createTypeAdapterSpec(@NotNull TypeTokenConstantsGenerator typeTokenConstantsGenerator,
```
yasirmhd: Yes, this will come in a very remote scenario. I think for now we can just remove this code.

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/model/AnnotatedClass.java
```diff
@@ -68,8 +63,10 @@
         FieldOption fieldOption = useStag != null ? useStag.value() : null;
         if (fieldOption == null) {
             useStag = findParentUseStagAnnotation(element);
-            // TODO: Do not default to FieldOption.SERIALIZED_NAME, instead set to null and fail after removing @GsonAdapterKey 1/30/17 [AR]
-            fieldOption = useStag != null ? useStag.value() : FieldOption.SERIALIZED_NAME;
+            fieldOption = useStag != null ? useStag.value() : null;
+
+            // The field option should never be null
+            Preconditions.checkNotNull(fieldOption);
```
yasirmhd: For the base classes it can possibly not have the annotation. Isn't it?

anthonycr: Hmmm, thinking about this now, I believe so. I don't think we have a model class in the sample that has a inherits from a non-annotated class, otherwise compilation would fail. I am going to remove this non null assertion and add in a test case for this scenario. Perhaps AnnotatedClass is named incorrectly 🤔

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/model/AnnotatedClass.java
```diff
@@ -67,7 +66,7 @@
 
         FieldOption fieldOption = useStag != null ? useStag.value() : null;
         if (fieldOption == null) {
-            useStag = ElementUtils.findAnnotation(UseStag.class, element);
+            useStag = element.getAnnotation(UseStag.class);
```
anthonycr: cc @mcumings I think this is okay to remove now, do you agree? The tests and build are all passing with their removal

mcumings: I think it makes sense to remove.  Technically it could still be supported (i.e., a lack of FieldOptions specification could cause a search through parent class annotations) but I think that would be surprising behavior to the consumer.  👍 to remove from me.

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/AdapterGenerator.java
anthonycr: Okay so I went through the code and it looks like the `TypeTokenConstantsGenerator` code should never be called. This is because in all the cases where it would have been called it would have been called in a case where a type was not a generic type (`TypeKind.TYPEVAR`) or did not contain any unresolved generic types in its declaration (e.g. `HashMap<String, T>`), but everywhere it was called was protected by either `TypeUtils.isConcreteType` or `TypeUtils.containsTypeVarParams` checks that would functionally prevent those branches from ever being hit. What I did was remove all the unnecessary code and add in preconditions checks to ensure that the conditions that are met right now in the code are reconsidered before the code is misused elsewhere. I also renamed some methods to specify that they are for generic types only.

See this commit: https://github.com/vimeo/stag-java/pull/77/commits/2a32acdf44a7ceb8b90cd69766e099cfad1863a2

yasirmhd: Looks good

mcumings:
@anthonycr QQ:  When do you anticipate being able to build a new release to incorporate this and the other recent changes?

anthonycr:
@mcumings I don't have a set date yet. There's a bug in our client app that may rely on https://github.com/vimeo/stag-java/issues/47. I'm working on that bug this week, so when I do I'll be able determine if that issue needs to be resolved or not before the next release (2.1.0).

mcumings:
Okay, thanks for the update.

Issue #81, opened 2017-04-07
Enum fields can't be private or final


anthonycr:
#### Issue Summary
In the 2.1.0 release, which more explicitly messages the consumer on when fields are private and/or final, a bug was introduced which prevents enum fields from being private or final. The cause of this bug was a global check for private and final fields on any class annotated with `@UseStag`. The correct solution would be to not impose those checks on enums, since enums are constant and don't need to serialize the fields.

#### Reproduction Steps
Create an enum that takes a parameter in its constructor. Make the field that it maps to private and final.

#### Expected Behavior
The Stag class to compile successfully.

#### Actual Behavior
The compilation fails due to the private and final fields.

anthonycr:
Fixed by https://github.com/vimeo/stag-java/pull/82

PR #82, opened 2017-04-10
Fix for enums not being able to have private or final fields.

anthonycr:
#### Ticket
https://github.com/vimeo/stag-java/issues/81

#### Ticket Summary
Enums could not have private or final fields as of the 2.1.0 release due to a bug in the modifier checking.

#### Implementation Summary
The solution was to ignore fields for enum types. A method was added to `TypeUtils` to determine whether an object was an enum type or not, and a test was added to verify it. A model has been added to the `sample-model` module to test this scenario.

#### How to Test
Run the tests.


kvenn:
Nice tests 👍  MWR

PR #110, opened 2017-06-01
Bumped up GSON version, performance fixes in JsonReader

anirudhramanan:
Refer https://github.com/google/gson/issues/1030


anthonycr:
👍

PR #113, opened 2017-06-15
Adding codecov.io integration

mikew-personal:
#### Ticket
N/A

#### Ticket Summary
I added this repo to our codecov.io account.  This change to the TravisCI yml file integrates their bash script to collect our test data.

#### Implementation Summary
- .travis.yml - Added an after_success action to call the codecov.io bash script.

#### How to Test
- This PR should show code coverage information if the integration is working properly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,8 @@ jdk:
 android:
   components:
     - tools
-    - build-tools-25.0.2
-    - android-25
+    - build-tools-26.0.0
+    - android-26
     - extra-android-support
     - extra-android-m2repository
 before_cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,8 @@ script:
 - rvm use 2.2
 - bundle install
 - FASTLANE_SKIP_UPDATE_CHECK=1 bundle exec fastlane test
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
 
 # https://vimeo.slack.com/services/B3TG0UCR3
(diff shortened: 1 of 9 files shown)


anthonycr:
If my memory from my previous experience with codecov is correct, we need to actually generate the jacoco test report during the build that will get picked up by codecov. This is because Android Studio is able to generate coverage reports itself, but that is not built into the gradle build system. We need to include additional dependencies and build setup.

I based my setup on how this repository (https://github.com/ReactiveX/RxJava) used codecov, and I arrived at the following setup:
1. Adding the jacoco plugin to the top level build.gradle: https://github.com/anthonycr/Bonsai/blob/dev/build.gradle#L10
2. Enabling code coverage for each build type in each module gradle build file: https://github.com/anthonycr/Bonsai/blob/dev/library/build.gradle#L20
3. <strike>Specifying some jacoco options if the tests utilize robolectric (stag's do not): https://github.com/anthonycr/Bonsai/blob/dev/library/build.gradle#L27</strike>
4. Specifying the jacoco version (this may be optional): https://github.com/anthonycr/Bonsai/blob/dev/library/build.gradle#L38
5. Running jacoco test report after all builds have been run: https://github.com/anthonycr/Bonsai/blob/dev/.travis.yml#L25

This seems like it might be a lot of work to do since stag has a lot of modules, so let me know if you want me to take on the work of integrating the jacoco stuff (if it truly is needed).

kvenn:
Oh man! Code cov!

Issue #112, opened 2017-06-14
Stag.Factory can only be used with a single Gson instance


mcumings:
#### Issue Summary

Stag.Factory is stateful which can cause problems when attempting to use a Stag.Factory
instance across multiple GsonBuilder invocations.

#### Reproduction Steps

Create a Stag.Factory instance A.
Create a TypeAdapter implementation B mapping java.util.Date.
Create a TypeAdapter implementation C mapping java.util.Date.
Create a type D to marshal that has a java.util.Date field in it.

Create a Gson instance using a builder configured with A and B.
Create a Gson instance using a builder configured with A and C.

The first Gson instance used to marshal/unmarshal type D will set a private field in A to cache the TypeAdapter instance.  If the second Gson instance is then used then the type adapter lookup is skipped in favor of the previously cached instance coming from the alternate Gson instance.

#### Expected Behavior

The current behavior is an implicit result of performance optimizations.  Given that reuse of a
Stag.Factory instance is a corner case the current behavior is probably reasonable.  I'm raising this
as an issue to document and highlight this surprising result to see if we want to do anything
about it.  For instance, it would be possible to capture the initial Gson instance used and then assert instance equality for subsequent uses rather than fail later in the process.

#### Actual Behavior

Described above.

anthonycr:
Added a test case in https://github.com/vimeo/stag-java/tree/bug/stateful-factory

anthonycr:
I'm thinking maybe we should just throw an exception if you try to reuse a factory with multiple gson instances. That seems like the best approach, what are your thoughts? What I would do is have the following in every method of the factory:

```java
private Gson mGson;

TypeAdapter<Blah> someMethodThatReturnsTypeAdapterBla(Gson gson) {
    setOrThrow(gson);

    return blah;
}

setOrThrow(Gson gson) {
    if (mGson != gson) {
        if (mGson != null) {
            throw new UnsupportedOperationException("This factory can only be used with a single Gson instance, please create a new instance");
        }

        mGson = gson;
    }
}
```

Now on the other hand, we could fix it so that it works with multiple instances, and I see two obvious options for allowing a factory to be used by multiple gson instances, but they both seem bad:
- Save off the gson instance when the `getMyClassTypeAdapter` method gets called, and if they are not the same, return a new type adapter. There would be a big problem performance-wise though, because if you are using two different gson instances with two different type adapters for a certain type simultaneously, then the gson instance would be changing all the time and it would always be recreating the type adapter. 
[...]

PR #115, opened 2017-06-19
[BUG] Stag.Factory does not support reuse and should tell the consumer so

anthonycr:
#### Ticket
https://github.com/vimeo/stag-java/issues/112

#### Ticket Summary
JSON was not properly parsed when trying to reuse a stag.factory instance between gson instances.

#### Implementation Summary
The chosen solution was to throw an `UnsupportedOperationException` when the consumer tries to reuse a `Stag.Factory` between `Gson` instances. See the linked ticket for more discussion. Due to the nature of this change, I am proposing that the next version of stag be `2.3.0` rather than `2.2.1`. The reasoning is that this could cause crashes for consumers that are reusing the factory between instances. In the worst case, consumers unaffected by the above issue (due to never registering different factories) will have to change their code to use a new instance instead of reusing it. I believe this is acceptable enough to do, let me know if you disagree.

#### How to Test
Run the added unit test to ensure that an exception is thrown. Note that the unit test will be able to test equality in a future where factory instances can be reused.


sample-java-model/src/test/java/com/vimeo/sample_java_model/SwappableParserExampleModelTest.java
```diff
@@ -0,0 +1,123 @@
...
+        SwappableParserExampleModel model1 = gson1.fromJson(typeAdapter1Json, SwappableParserExampleModel.class);
+
+        Assert.assertEquals(model1.testField2.testField, "test");
+
+        // TypeAdapter2 assumes a reversed string
+        String typeAdapter2Json = swappableParserExampleJsonWithTestObjectField("tset");
+
+        Gson gson2 = new GsonBuilder().registerTypeAdapterFactory(factory).registerTypeAdapter(TestObject.class, new TestObjectAdapter2()).create();
+
+        SwappableParserExampleModel model2 = gson2.fromJson(typeAdapter2Json, SwappableParserExampleModel.class);
```
msya: `typeAdapter1Json`, `gson2`and `model2` could be `final`.

sample-java-model/src/main/java/com/vimeo/sample_java_model/SwappableParserExampleModel.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A model to test the case where the consumer
+ * might try to use different parsers for
+ * an object across different gson instances.
+ */
+@UseStag
+public class SwappableParserExampleModel {
```
msya: It seems this file is only used for unit tests. Should it be in the test package? Should this package `private` rather `public`?

sample-java-model/src/main/java/com/vimeo/sample_java_model/SwappableParserExampleModel.java
```diff
@@ -0,0 +1,28 @@
...
+
+    @NotNull
```
msya: There are lint warning for the `@NotNull` annotations that they must be initialized.

sample-java-model/src/main/java/com/vimeo/sample_java_model/SwappableParserExampleModel.java
```diff
@@ -0,0 +1,28 @@
...
+    String testField1;
+
+    @NotNull
+    TestObject testField2;
+
+    public static class TestObject {
```
msya: Should this be package private?

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/StagGenerator.java
```diff
@@ -299,8 +299,14 @@ private TypeSpec getAdapterFactorySpec() {
                 .addParameter(Gson.class, "gson")
                 .addParameter(ParameterizedTypeName.get(ClassName.get(TypeToken.class), genericTypeName),
                               "type")
+                .addStatement("setOrThrow(gson)")
```
msya: There is a license JavaDoc in this file. But, it is not in other files. Should it be removed?

sample-java-model/src/main/java/com/vimeo/sample_java_model/SwappableParserExampleModel.java
anthonycr: I'm going to leave it in the same package for now (because the main reason for this module is just for testing), but good point on making it package private.

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/StagGenerator.java
anthonycr: There should be a license header on every file, so I will add it where missed. Good catch.

sample-java-model/src/main/java/com/vimeo/sample_java_model/SwappableParserExampleModel.java
anthonycr: Actually, I have to take that back. They have to be public because the generated Stag.Factory that references both this class and the inner class is being generated in `com.vimeo.sample_java_model.stag.generated` which is outside the package.

anthonycr:
@msya responded to PR comments https://github.com/vimeo/stag-java/pull/115/commits/b0e4c5e5d08cc1d06db16a73244d213288f1876b

Issue #120, opened 2017-07-12
"Cannot find getter or setter" for private parameterized member variables


anthonycr:
#### Issue Summary
Given the following code
```java
private ArrayList<String> myList;

public void setMyList(ArrayList<String>) { ... }

public ArrayList<String> getMyList() { ... }
```

Stag will be unable to find the setters. Only parameterized fields are affected by this bug.

#### Reproduction Steps
See above

#### Expected Behavior
It should find the getter/setter.

#### Actual Behavior
The build fails.

The cause of this bug is incorrect comparison of `TypeMirror` objects, using the correct comparison will fix the problem.

anthonycr:
Closing as fixed.

PR #121, opened 2017-07-12
Fix bug where private parameterized fields could not be found

anthonycr:
#### Ticket
https://github.com/vimeo/stag-java/issues/120

#### Ticket Summary
See https://github.com/vimeo/stag-java/issues/120

#### Implementation Summary
I replaced usage of `typeMirror1.equals(typeMirror2)` with `Types.isSameType(typeMirror1, typeMirror2)` as is suggested in this OpenJDK issue https://bugs.openjdk.java.net/browse/JDK-6498938

I also modified the `sample-java` model to handle this test case.

I also improved error messaging around the private member variable compilation failures.

#### How to Test
Run the tests.

### DIFF
--- a/sample-model/build.gradle
+++ b/sample-model/build.gradle
@@ -21,8 +21,9 @@ android {
         javaCompileOptions {
             annotationProcessorOptions {
                 arguments = [
-                        stagGeneratedPackageName: 'com.vimeo.sample_model.stag.generated',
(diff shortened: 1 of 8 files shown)


anthonycr:
@msya I'll work on cutting a release with these changes so you're unblocked

PR #125, opened 2017-08-14
Updating processor argument example

anthonycr:
#### Summary

The java example in the readme for passing processor arguments to the compiler was wrong. I updated the readme example to pass arguments to the compiler using the gradle plugin as it should. I also updated the java example module to use this same format instead of using the manual way.

### DIFF
--- a/sample-model/build.gradle
+++ b/sample-model/build.gradle
@@ -58,6 +58,9 @@ dependencies {
 
 gradle.projectsEvaluated {
     tasks.withType(JavaCompile) {
-        options.compilerArgs << "-Xlint:all,-deprecation,-serial,-processing" << "-Werror"
+        options.compilerArgs += [
+                "-Xlint:all,-deprecation,-serial,-processing",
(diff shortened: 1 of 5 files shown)


sample-model/build.gradle
```diff
@@ -58,6 +58,9 @@ dependencies {
 
 gradle.projectsEvaluated {
     tasks.withType(JavaCompile) {
-        options.compilerArgs << "-Xlint:all,-deprecation,-serial,-processing" << "-Werror"
+        options.compilerArgs += [
```
brentwatson: Not sure if you want to update the samples to use v0.11 of ltgt.

anthonycr: Only `sample-java-model` is utilizing the ltgt plugin (it's on the latest), the rest use the android annotation processor stuff, this block just turns on java lint all the way up and sets warnings as errors so that any bad generated code causes the build to fail 👍

Issue #132, opened 2017-09-18
Generated TypeAdapter writes out an empty object for null values


mcumings:
#### Issue Summary

When the generated TypeAdapter is instructed to write a `null` value it always writes an empty object.  This is due to the following code at the beginning of the `write(JsonWriter, T)` method:
```
    writer.beginObject();
    if (object == null) {
      writer.endObject();
      return;
    }
``` 

#### Reproduction Steps

Create a type adapter and use it to write a `null` value.

#### Expected Behavior

The generated TypeAdapter should wither write out a JSON null in this scenario or should be configurable to allow either behavior at the discretion of the user.

#### Actual Behavior

A JSON Object with no fields is written.

mcumings:
Raising this for discussion as to how we should handle this.  Our app has code which expects to write a JSON null in this scenario.

We could add an option to @UseStag to allow the user to declare the desired behavior, defaulting operation to writing an empty object.  Any opinion on this?

anthonycr:
@mcumings I think this falls squarely in the realm of a bug. I can't think of any reason why it would be advantageous to write an empty object rather than a null object in this scenario, and I would assume this would cause problems for parsing in most cases where a field could be null, but the fields in the object are annotated `@NotNull` or something and therefore an object which succeeds to be written to JSON would fail to be read from the same JSON.

Unless anyone else thinks otherwise, I would advocate for just switching the behavior completely and putting the change behind a minor version bump.

anthonycr:
Another bug I found, if you have an array with a null value, it will emit an empty array instead of null. I am fixing those as well.

anthonycr:
See: https://github.com/vimeo/stag-java/pull/133

mcumings:
Sorry for the delay in getting back to this and thanks for already taking a stab at it!

I think we're safe after this change with respect to the ingress of nulls.  I think your final suggestion relates only to writing the null values in which case there are only two scenarios that I can think of:

1. The root object or array being marshalled is null.  I think we have no choice other than to write the null there.  This is the scenario being handled in this issue.
1. A non-null object is attempting to write a null field.  Currently this no-ops and skips that field.  I think this is the only point where additional configuration could be applied to select whether or not to actually write a null value versus just skipping it.

[...]

PR #133, opened 2017-10-02
Fixing empty object for null values

anthonycr:
#### Issue
- https://github.com/vimeo/stag-java/issues/132

#### Summary
Previously, when stag encountered a null value when writing objects to json, it would write an empty object. This behavior is erroneous because a field could be annotated `Nullable`, but the field type could have `NonNull` fields. This means that the following valid objects could cause exceptions to be thrown when reading the object out of JSON:

```java
class NullableFieldExample {
    @Nullable public MyObject myObject;
}

class MyObject {
    @NonNull public String title;
}
```
For the following object
```java
NullableFieldExample object = new NullableFieldExample();
object.myObject = null;
gson.toJson(object);
```
the following JSON will be produced.
```json
{
"myObject": {}
}
```
If we attempt to read this object out of JSON, we will get an exception because the field `title` will be null.

Based on the implications of these changes, I would suggest that the next version of stag receive a minor version bump to 2.4.0.

#### How to Test
Test examples, run tests.

### DIFF
--- a/sample-java-model/src/main/java/com/vimeo/sample_java_model/NullFields.java
+++ b/sample-java-model/src/main/java/com/vimeo/sample_java_model/NullFields.java
@@ -1,0 +1,36 @@
+package com.vimeo.sample_java_model;
+
+import com.google.gson.annotations.SerializedName;
+import com.vimeo.stag.UseStag;
+
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A test case for an object containing null fields.
+ * <p>
+ * Created by restainoa on 10/20/17.
+ */
+@UseStag
+public class NullFields {
+
+    @Nullable
+    @SerializedName("hello")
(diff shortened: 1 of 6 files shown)


anthonycr:
Assigning @howardrigberg and unassigning @msya as I have given him the other PR.

mcumings:
Can we get some tests for these scenarios to guard them moving forward?

howardrigberg:
I will get to this tomorrow.

anthonycr:
@howardrigberg a test case has been added

mcumings:
👍 

Thanks a ton for jumping on this.  If I wasn't buried at the moment I would have happily contributed a patch but wasn't able to beat you to it. 😄

anthonycr:
Merging. Going to work on a 2.4.0 release with this stuff in it.

PR #134, opened 2017-10-02
Testing type adapters created in kotlin module

anthonycr:
#### Issue
Currently, we are not testing the generated type adapters and are just testing that the type adapters themselves have been generated. In order to verify that the code we are generating is correct, we need to test each type adapter.

#### Summary
This PR brings the first type adapter testing to the library. Using a [library](https://github.com/devopsfolks/podam) called [PODAM](https://devopsfolks.github.io/podam/), we can fill POJOs with data. Using these filled POJOs, we can generate JSON using the generated type adapter. We then recreate the object from the JSON using the same type adapter, and then we compare the objects, asserting that they are the same.

PODAM works by calling setters on an object, and due to its strict naming policy, it doesn't pick up fields named using hungarian notation, nor boolean fields using the `isXYZ -> isXYZ()` naming convention. Otherwise, it picks up and sets all fields.

Going forward, there should be some sort of strategy in place for converting existing POJOs away from hungarian notation, OR we should write our own `ClassInfoStrategy` and set it on the factory, which allows PODAM to find these fields.

#### How to Test
Run the modified tests and verify that they work.

- [x] Switch branch to develop after https://github.com/vimeo/stag-java/pull/133 is merged


anthonycr:
@msya this needs review.

sample-kotlin-model/src/main/kotlin/com/vimeo/dummy/sample_kotlin/BoolFields.kt
```diff
@@ -9,12 +9,7 @@ import com.vimeo.stag.UseStag
...
-    @SerializedName("test1")
-    var test1: Boolean? = null
-
-    @SerializedName("test2")
-    var isTest2: Boolean? = null
-
-}
\ No newline at end of file
+data class BoolFields(
+        @SerializedName("test1") var test1: Boolean? = null,
```
msya: Why does this field require a `SerializedName` annotation with the same name as the variable?

sample-kotlin-model/src/main/kotlin/com/vimeo/dummy/sample_kotlin/KotlinSamples.kt
```diff
@@ -9,17 +9,10 @@ import com.vimeo.stag.UseStag
...
-
-    var testField: Any? = null
-}
\ No newline at end of file
+data class KotlinSamples(
+        var stringField: String? = null,
+        var nonNullStringField: String = "default",
+        var intField: Int? = null,
+        var longField: Long = 1,
+        @SerializedName("boolean_field") var booleanField: Boolean? = null
```
msya: In this data class, other fields are not annotated. Why does only this field require annotation?

sample-kotlin-model/src/main/kotlin/com/vimeo/dummy/sample_kotlin/KotlinConcreteExample.kt
```diff
@@ -12,4 +12,22 @@ class KotlinConcreteExample : KotlinGenericExample<String>() {
 
     var kotlinObject: KotlinSamples? = null
 
+    override fun equals(other: Any?): Boolean {
```
msya: Are `sample-kotlin-model` and `sample-java-model` library only related modules or they are examples for the end user?

anthonycr: They are primarily used as test cases, but they also serve as examples. In the future I will be moving all the "samples" into an integration test module, and will leave the sample project as just a sample.

sample-kotlin-model/src/main/kotlin/com/vimeo/dummy/sample_kotlin/BoolFields.kt
anthonycr: It does not require the annotation, it is just added for consistency.

sample-kotlin-model/src/main/kotlin/com/vimeo/dummy/sample_kotlin/KotlinSamples.kt
anthonycr: annotating fields is optional, however, I will add the annotation to the other fields since it is inconsistent here.

Issue #142, opened 2017-10-26
Custom TypeAdapterFactory not called when using Stag.Factory


ursusursus:
Hi,
I have a custom PostParseTypeAdapterFactory , which just calls onParseCompleted method on my objects after parsing is finished, to allow for some post processing of values, etc.

```
public class PostParseTypeAdapterFactory implements TypeAdapterFactory {
    public interface PostProcessable {
        void onParseCompleted();
    }

    @Override public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
        final TypeAdapter<T> delegate = gson.getDelegateAdapter(this, type);
        return new TypeAdapter<T>() {
            public void write(JsonWriter out, T value) throws IOException {
                delegate.write(out, value);
            }

            public T read(JsonReader in) throws IOException {
                T obj = delegate.read(in);
                if (obj instanceof PostProcessable) {
                    ((PostProcessable) obj).onParseCompleted();
                }
                return obj;
            }
        };
    }
}
```

When not using Stag factory, it works as expected, **however with it, the onParseCompleted doesnt get called**

anthonycr:
I'll look into this. 

As a work around, if you are able to edit the `PostParseTypeAdapterFactory`, you can initialize it with a `Stag.Factory` instance:
```java
public class PostParseTypeAdapterFactory implements TypeAdapterFactory {

        public interface PostProcessable {

            void onParseCompleted();
        }

        private final TypeAdapterFactory delegateFactory;

        public PostParseTypeAdapterFactory(final TypeAdapterFactory delegate) {
            delegateFactory = delegate;
        }

        @Override
        public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
            final TypeAdapter<T> delegate = delegateFactory.create(gson, type);
            return new TypeAdapter<T>() {
                public void write(JsonWriter out, T value) throws IOException {
                    delegate.write(out, value);
                }

                public T read(JsonReader in) throws IOException {
                    T obj = delegate.read(in);
                    if (obj instanceof PostProcessable) {
                        ((PostProcessable) obj).onParseCompleted();
                    }
                    return obj;
                }
            };
        }
    }

// Use as follows
final Gson gson = new GsonBuilder()
                        .registerTypeAdapterFactory(new PostParseTypeAdapterFactory(new Stag.Factory()))
                        .create();
```

ursusursus:
[...]

PR #145, opened 2017-11-03
Cleaner Stag.Factory, and performant way of using TypeToken

anirudhramanan:
**Changes:** 

**Removal of TypeAdapter member variables, and getter methods from Stag.Factory :**

The Stag.Factory used to get overloaded with 700+ TypeAdapter getter methods, and member variables (in our use-case) due to which the class loading takes time. When we profiled we came to know that it was taking ~400ms just for loading the factory class. 

**Performant way of using TypeToken:**

There are two ways of initialising type tokens. 
1.  using the constructor, and 
2. using TypeToken.get(type). 

The first one uses reflection when it tries to get its generic superclass, and it is slower as compared to the second one.

In case of ParameterizedTypes as well, there is a method exposed in the TypeToken class `TypeToken.getParameterized(type, typearguments...)` which is way efficient than creating typetoken using the constructor


anthonycr:
So is this a work in progress with the intent to merge in the future? Or is help needed? Wondering whether I should wait provide feedback or not.

yasirmhd:
@anthonycr , kindly review this and give us feedback. We are going to run this through more testing before we would like this to be merged. At this stage, we will require only help with code reviews and your thoughts/insights on this approach.

**Problem Statement :**
Stag.Factory class was growing. In our case as @anirudhramanan mentioned, it had gone huge with just loading this taking upwards of 400 ms (it had to load all the classes). We had to actually put the initialisation of this class on a background thread. It is impacting our cold startup time, as before we show something meaningful which the user can interact with. 

The intent is to give a more scalable way of doing this, and make the Stag Factory light enough so that it can be initialised on the main thread. 


**Solution :**  
The idea behind this came when we raised this [PR](https://github.com/google/gson/pull/1064) in GSON code. 

TypeToken is expensive when you use the default constructor (use it as an anonymous class). Using the constructor which uses the Type as a variable is not expensive. 

Given that we are doing code generation, we do know the exact types of the member variables of a class and hence we can easily construct TypeToken objects which do not use the default constructor (even for types which are generic/parametrised) we can infer the type variables and create them. 

Here is what is being done here : 

1. Move to creating TypeTokens in the adapter themselves. Keep the Adapters as member variables and initialize them in the constructor. Given that Gson.getAdapter caches the adapters by TypeTokens, we will not be actually creating an Adapter every time and they will still be re-used.

2. Create TypeTokens for non-generic classes and keep  them as public static final member variables. This ensures that we are also optimizing on creation of TypeTokens as much as we can. This also ensures the distribution of the variables and they are not kept in a single class or file, which can again lead to class loading slowness. 

3. There continue to be two exceptions here. Map, Collections and primitive types. Here is the reason for that 
      
        -   Maps and Collections : Needed a way to insitantiate these objects while reading it, so had to do it. Have distributed the impact by creating an adapter in the class where they are being used. The TypeAdapter class for this is not very expensive to create so it should be OK
        - Primitives : We saw earlier in one of the PRs that avoiding boxing and unboxing led to huge perf gains. So we are sticking with that.

yasirmhd:
Update : we are doing the profiling. The results right now point to some gains, but the Stag.Factory class is still huge. We have some solution in mind. Will get back on this by Thursday

anthonycr:
Sounds good to me

anthonycr:
Interestingly, this recently opened bug is related to this PR: https://github.com/vimeo/stag-java/issues/142

yasirmhd:
@anthonycr  : Yes, it will fix issue #142. Overall, this seems like a way of doing things which are closer to Gson. There are some gains that we have seen with this approach. However we still need to make it perform better.

Unfortunately, I have not been able to get back to this for a week. Hopefully by next week i should get some time to spend on this.

anirudhramanan:
@anthonycr The PR is ready to be reviewed. 

With these changes, each package will generate it's own StagFactory, and the main Stag.Factory in the generated package will loop through all the sub factories to return the correct one. We did some profiling with these changes, and got a huge improvement in the class load time. 

We tested this on a project with appox ~900 pojos.

----OLD STAG LOADING TIME----
Stag.Factory : 245970, 175353, 172015 : 197,779 (197.779 ms)

----NEW STAG LOADING TIME----
Stag.Factory : 10920, 11240, 10914 : 11024 (11.024 ms)

The Stag.Factory now looks cleaner with only create() method.

anthonycr:
I'll begin reviewing this today

yasirmhd:
We are running into this GSON performance issue with this approach. We have submitted a [PR](https://github.com/google/gson/pull/1199) with GSON. We will probably have to wait for the comments there or look at an alternate approach.

cc: @anthonycr

anthonycr:
Latest testing changes have created merge conflicts in the models, looks like all packaging changes. Sorry for the inconvenience caused there. Just to explain the way the packages are structured now to help with resolving the conflicts, there are now 5 non-library modules:
- integration-test-android
- integration-test-java-cross-module
- integration-test-java
- integration-test-kotlin
- sample

The restructuring was done to make the library source code more understandable and allow the test modules to be pulled into `stag-library-compiler` as test cases. The way tests will work from now on is that "integration" tests will exist in all modules named "integration-" and `sample` will also have integration tests. All the remaining tests in these modules are integration tests that provide no test coverage on the library itself, but test the generated code. The `integration-test-java-*` modules are pulled into `stag-library-compiler` as test cases for "functional" tests, which assert that compilation succeeded or failed for certain inputs. Unit tests exist solely in `stag-library-compiler` now.

`sample` contains no test cases now (other than some integration tests), and should just be used for the actual android example code.

`integration-test-android` only provides test cases for models which contain platform types. Models should not be added to this module unless they are dependent on the Android OS to operate. Not pulled in by the compiler module as it relies on platform types and therefore this module does not provide test coverage for the library.

`integration-test-java` provides test cases written in java. All new java test cases should go into this module.

`integration-test-java-cross-module` provides test cases for models which rely on other models from `integration-test-java`. It also contains models which should be in `integration-test-java` because they do not rely on types from another module, but I did not move them yet because I didn't want to make the conflicts any worse.

`integration-test-kotlin` provides test cases written in Kotlin. Currently not pulled in by the compiler module due to google's `compile-testing` library not being able to compile kotlin.

anirudhramanan:
@anthonycr np, will rebase it with dev. Also we have pushed some more changes today which gave has some significant performance improvements in the getAdapter call.

anirudhramanan:
@anthonycr We profiled the cold startup of the app, and were able to shave off ~200 ms on the startup time. The StagFactory class loading does not take much time now as compared to the previous versions. Since we have approx ~900 models, the StagFactory now contains approx ~1200 lines of code as compared to ~10000 lines. This has helped us in improving the overall cold startup time.

The PR is ready to be reviewed.

yasirmhd:
@anthonycr  : We are good to merge this from our end now. This is giving us better benefits and has improved the overall deserialisation time as well. 

What we have also done in this PR is try to get to the sub factory on the basis of the package of the class. We have do to this check at run time because code obfuscation would mean that the package and class name will not be the same as was when we generated the code for the factory. Given the possible number of sub factories it is also not optimal to preload all of them as the performance profiling showed us that it would load all of these factory class and slow down the cold boot times for the app.

The Google PR is something which will give us further benefits, but at this time it seems to be better in performance from the existing implementation already.

yasirmhd:
@anthonycr  Travis seems to be breaking because of some issues regarding license agreement. We are trying to check what could have caused this because of this PR. 

In the meanwhile we will really appreciate if you can give your feedback on this PR.

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/AdapterGenerator.java
```diff
@@ -77,4 +84,12 @@ static String getJsonName(@NotNull Element element) {
      */
     @NotNull
     public abstract TypeSpec createTypeAdapterSpec(@NotNull StagGenerator stagGenerator);
+
+
+    FieldSpec createTypeTokenSpec(@NotNull TypeMirror typeMirror){
```
anthonycr: Since this is part of the public (well package local) API of this class, it would be good to have javadocs on it.

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/StagFactoryGenerator.java
```diff
@@ -0,0 +1,94 @@
...
+    public static final String NAME = "StagFactory";
+    private final List<ClassInfo> classInfoList;
+    private final String fileName;
+
+    public StagFactoryGenerator(List<ClassInfo> classInfoList, String fileName) {
+        this.classInfoList = classInfoList;
+        this.fileName = fileName;
+    }
+
+    public TypeSpec getTypeAdapterFactorySpec() {
```
anthonycr: annotate NotNull

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/AdapterGenerator.java
anthonycr: Also could be annotated NotNull

stag-library-compiler/src/main/java/com/vimeo/stag/processor/StagProcessor.java
```diff
@@ -169,10 +214,20 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
         return true;
     }
 
+    private void generateAdapterFactory(List<ClassInfo> classInfos, String packageName) throws IOException {
```
anthonycr: minor: params should be annotated NotNull

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/StagFactoryGenerator.java
```diff
@@ -0,0 +1,94 @@
...
+
+import javax.lang.model.element.Modifier;
+import javax.lang.model.type.TypeMirror;
+
+public class StagFactoryGenerator {
+    public static final String NAME = "StagFactory";
+    private final List<ClassInfo> classInfoList;
+    private final String fileName;
+
+    public StagFactoryGenerator(List<ClassInfo> classInfoList, String fileName) {
```
anthonycr: Parameters should be annotated NotNull.

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
```diff
@@ -866,6 +595,18 @@ public TypeSpec createTypeAdapterSpec(@NotNull StagGenerator stagGenerator) {
         return adapterBuilder.build();
     }
 
+    private static class FieldInfo {
+        private final TypeMirror type;
+        private final String initializationCode;
+        private final String accessorVariable;
+
+        FieldInfo(TypeMirror type, String initializationCode, String accessorVariable) {
```
anthonycr: Add nullability annotations here

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
```diff
@@ -866,6 +595,18 @@ public TypeSpec createTypeAdapterSpec(@NotNull StagGenerator stagGenerator) {
         return adapterBuilder.build();
     }
 
+    private static class FieldInfo {
+        private final TypeMirror type;
+        private final String initializationCode;
+        private final String accessorVariable;
```
anthonycr: Add nullability annotations to these fields

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
```diff
@@ -90,135 +84,46 @@ public TypeAdapterGenerator(@NotNull SupportedTypesModel supportedTypesModel, @N
...
+    private static String getTypeTokenCode(@NotNull TypeMirror fieldType,
+                                           @NotNull StagGenerator stagGenerator,
+                                           @NotNull Map<TypeMirror, String> typeVarsMap,
+                                           @NotNull AdapterFieldInfo adapterFieldInfo) {
         if (fieldType.getKind() == TypeKind.TYPEVAR) {
-            result = "com.google.gson.reflect.TypeToken.get(" + typeVarsMap.get(fieldType) + ")";
+            return adapterFieldInfo.updateAndGetTypeTokenFieldName(fieldType, "(com.google.gson.reflect.TypeToken<" + fieldType.toString() + ">) com.google.gson.reflect.TypeToken.get(" + typeVarsMap.get(fieldType) + ")");
+        } else if (!TypeUtils.isParameterizedType(fieldType)) {
+            ClassInfo classInfo = stagGenerator.getKnownClass(fieldType);
+            if(classInfo != null) {
```
anthonycr: formatting problems in this file, please run the autoformatter on this

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
```diff
@@ -90,135 +84,46 @@ public TypeAdapterGenerator(@NotNull SupportedTypesModel supportedTypesModel, @N
...
-            result = "com.google.gson.reflect.TypeToken.getParameterized(" +
-                     declaredFieldType.asElement().toString() + ".class";
+            String result = "(com.google.gson.reflect.TypeToken<" + fieldType.toString() + ">)com.google.gson.reflect.TypeToken.getParameterized(" +
+                    declaredFieldType.asElement().toString() + ".class";
                 /*
                  * Iterate through all the types from the typeArguments and generate type token code accordingly
                  */
             for (TypeMirror parameterTypeMirror : typeMirrors) {
-                if (TypeUtils.isSupportedNative(parameterTypeMirror.toString())) {
+                if(parameterTypeMirror.getKind() != TypeKind.TYPEVAR && !TypeUtils.isParameterizedType(parameterTypeMirror)) {
```
anthonycr: formatting problems in this file, please run the formatter here.

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
```diff
@@ -348,324 +262,147 @@ private static String getFieldAccessorForKnownJsonAdapterType(@NotNull Executabl
...
-                sStagFactoryUsed = true;
-                String adapterCode = "new " + TypeUtils.className(ObjectTypeAdapter.class) + "(mGson)";
-                String getterName = stagGenerator.addFieldForKnownType(fieldType,
-                                                                       adapterCode.replaceAll("mStagFactory.",
-                                                                                              "")
-                                                                               .replaceAll("mGson", "gson"));
-                return "mStagFactory." + getterName + "(mGson)";
-            } else if (fieldType instanceof DeclaredType) {
+        String fieldName = adapterFieldInfo.getFieldName(fieldType);
+        if(null != fieldName) {
```
anthonycr: formatting problem, could you run the formatter on this file?

stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/AdapterGenerator.java
anthonycr: Could you run the formatter on this file?

anirudhramanan:
@anthonycr Thanks for the review. Will make the required changes by today

build.gradle
```diff
@@ -8,7 +8,7 @@ buildscript {
     // android dependencies
     ext.targetSdk = 27
     ext.minSdk = 14
-    ext.buildTools = "27.0.1"
+    ext.buildTools = "27.0.2"
```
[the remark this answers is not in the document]
anirudhramanan: working now

PR #150, opened 2017-12-18
thread safe changes, moved inside the synchronized block.

anirudhramanan:


yasirmhd:
@anthonycr : kindly review this. This should be better implementation from a thread safety purpose. We are also running the profiling and will update the results in case we see anything adverse from this check in.

PR #152, opened 2017-12-19
Thread-Safety Changes, moved the getSubFactory to the synchronised block

anirudhramanan:
Earlier the entire logic of the type adapter creation was inside the synchronised block. In cases where two threads were to access it, one thread would have to wait until the 1st thread creates the TypeAdapter and returns. 

Now, with this change, the thread will wait only till it finds the correct subFactory and the typeadapterfactory.create() will happen outside the lock.

We now have another method `getSubFactory` which will return the correct sub factory for a given class. This method is synchronised. Once we get the correct factory, we call the `.create()` as usual and return the adapter


anthonycr:
I'll review this tomorrow

PR #154, opened 2018-01-03
Resolve #153 by moving suppressing across the entire generated TA class

### LINKED ISSUE #153: Fields which use raw types emit compilation warnings
#### Issue Summary

Model classes which refer to generically typed objects cause TypeAdapter references to be generated which cause "rawtypes" warnings to be generated within the generated TypeAdapter.

We have currently have suppressions for "rawtypes" and "unchecked" on the generated TypeAdapter's constructor but the failing point is in the generated TypeAdapter's `mTypeAdapterN` field references.

#### Reproduction Steps

Will include a test case in the PR.

#### Expected Behavior

We suppress the issues for field definitions as well.

#### Actual Behavior

The generated code causes javac lint "rawtypes" warnings.

mcumings:
If the solution is not applied then the RawGenericField class will cause
the compilation warning to be promoted to an error, failing the
integration-test-java module.

#### Issue

https://github.com/vimeo/stag-java/issues/153

#### Summary

Moves suppression annotations from the constructor up to the class level.

We could alternately add them to the field definitions but I fear we'll end up with these issues propagating to another location within the generated code if we find new ways to exercise this.

#### How to Test
Build the project and inspect com.vimeo.sample_java_model.RawGenericField$TypeAdapter source.

### DIFF
--- a/integration-test-java/src/main/java/com/vimeo/sample_java_model/RawGenericField.java
+++ b/integration-test-java/src/main/java/com/vimeo/sample_java_model/RawGenericField.java
@@ -1,0 +1,15 @@
+package com.vimeo.sample_java_model;
+
+import com.vimeo.stag.UseStag;
+
+/**
+ * Model which references a generically typed object without generic bounds (i.e., as a raw
+ * type).
+ */
+@UseStag
+public class RawGenericField {
+
+    @SuppressWarnings("rawtypes")
+    public ExternalModelGeneric rawTypedField;
+
+}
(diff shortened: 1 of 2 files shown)


mcumings:
There's also an issue with wildcards in generics.  I'm working up something for that as well (I hope).

mcumings:
@anthonycr - Doesn't look like anyone is getting automatically added to reviews so I'm manually mentioning you. 😄

anthonycr:
thanks for requesting me @mcumings, I forgot about this one. I'll give it a review today 👍

Issue #155, opened 2018-01-09
Fields referencing wildcarded generics cause compilation failure


mcumings:
#### Issue Summary

Defining models with wild-carded generics such as the following result in invalid Java sources being generated for the TypeAdapter.

This scenario is useful when consuming models of varying types into a single collection.  For example, the code may leverage a provided type adapter factory to dynamically determine how to unmarshal the a JSON object payload, dynamically resolving the correct TypeAdapter to use for the object type being unmarshaled.

#### Reproduction Steps

Define a model which uses wildcarded generics such as the following:

```
@UseStag
public class DynamicallyTypedWildcardReadModel {
    public List<DynamicallyTypedModel<?>> models;
}
```

The generated code for the failing constructor looks like this:

```
  public DynamicallyTypedWildcardReadModel$TypeAdapter(Gson gson) {
    this.mGson = gson;
    com.google.gson.reflect.TypeToken<com.vimeo.sample_java_model.DynamicallyTypedModel<?>> typeToken0 = (com.google.gson.reflect.TypeToken<com.vimeo.sample_java_model.DynamicallyTypedModel<?>>)com.google.gson.reflect.TypeToken.getParameterized(com.vimeo.sample_java_model.DynamicallyTypedModel.class, ?.class);
    this.mTypeAdapter0 = gson.getAdapter(typeToken0);
    this.mTypeAdapter1 = new com.vimeo.stag.KnownTypeAdapters.ListTypeAdapter<com.vimeo.sample_java_model.DynamicallyTypedModel<?>,java.util.List<com.vimeo.sample_java_model.DynamicallyTypedModel<?>>>(mTypeAdapter0, new com.vimeo.stag.KnownTypeAdapters.ListInstantiator<com.vimeo.sample_java_model.DynamicallyTypedModel<?>>());
  }
```

Which results in compilation failure:

```
/Users/mcumings/src/github/stag-java/integration-test-java/build/generated/source/apt/main/com/vimeo/sample_java_model/DynamicallyTypedWildcardReadModel$TypeAdapter.java:28: error: illegal start of expression
    com.google.gson.reflect.TypeToken<com.vimeo.sample_java_model.DynamicallyTypedModel<?>> typeToken0 = (com.google.gson.reflect.TypeToken<com.vimeo.sample_java_model.DynamicallyTypedModel<?>>)com.google.gson.reflect.TypeToken.getParameterized(com.vimeo.sample_java_model.DynamicallyTypedModel.class, ?.class);
[...]

PR #156, opened 2018-01-09
Adds tests and hacked solution for use case documented in #155

mcumings:
#### Issue

#### Summary

Commits tests which demonstrate the use case for supporting unmarshaling of wildcarded generics.

This code applies a hacky solution that creates a TypeToken
that *appears* to capture the wildcarded generics.  e.g.:

`new TypeToken<DynamicallyTypedModel<?>>() {};`

This explicitly violates the contract of TypeToken in its Javadoc:

`This syntax cannot be used to create type literals that have wildcard parameters, such as Class<?> or List<? extends CharSequence>`

It *appears* to work but is only committed as an initial step towards
finding a better solution.

#### How to Test

Tests included.

### DIFF
--- a/stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
+++ b/stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
@@ -110,7 +110,10 @@ public class TypeAdapterGenerator extends AdapterGenerator {
                  * Iterate through all the types from the typeArguments and generate type token code accordingly
                  */
             for (TypeMirror parameterTypeMirror : typeMirrors) {
-                if (parameterTypeMirror.getKind() != TypeKind.TYPEVAR && !TypeUtils.isParameterizedType(parameterTypeMirror)) {
+                if (parameterTypeMirror.getKind() == TypeKind.WILDCARD) {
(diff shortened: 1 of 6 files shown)


stag-library-compiler/src/main/java/com/vimeo/stag/processor/generators/TypeAdapterGenerator.java
```diff
@@ -110,7 +110,10 @@ private static String getTypeTokenCode(@NotNull TypeMirror fieldType,
                  * Iterate through all the types from the typeArguments and generate type token code accordingly
                  */
             for (TypeMirror parameterTypeMirror : typeMirrors) {
-                if (parameterTypeMirror.getKind() != TypeKind.TYPEVAR && !TypeUtils.isParameterizedType(parameterTypeMirror)) {
+                if (parameterTypeMirror.getKind() == TypeKind.WILDCARD) {
+                    return adapterFieldInfo.updateAndGetTypeTokenFieldName(fieldType,
+                            "new com.google.gson.reflect.TypeToken<" + fieldType + ">() {}");
+                } else if (parameterTypeMirror.getKind() != TypeKind.TYPEVAR && !TypeUtils.isParameterizedType(parameterTypeMirror)) {
```
mcumings: This is what allows it to compile but needs to be replaced prior to merge...

PR #157, opened 2018-01-09
#155 Flag wildcard generic usage as an error

mcumings:
#### Issue

Related: #156 

#### Summary

This PR presents code which takes the easy way out of this problem for the near term.  I'm posting it as an option for moving forward...

Until such time as direct support is added, consider it a compile-time error to attempt to use Stag with a model that references types containing wildcard generics.  This at least makes the failure mode explicit and communicates to the consumer what the underlying problem is.

#### How to Test

Tests included.


mcumings:
No longer necessary.

PR #158, opened 2018-01-10
#155 Rough implementation of wildcard generics forwarding to Gson

mcumings:
#### Issue

#### Summary

Identifies fields which leverage wildcarded generics and:
- Prevents them from generating mTypeAdapterN fields
- write calls uses the actual field type of the written object
- read calls use reflection to determine field type at runtime

This commit still needs work and testing.

#### How to Test

Test included.

Work needed:
- More tests covering field access in super-classes / sub-classes
- Possibly pull read and write- time TypeAdapter lookups up to the method level so that the lookups can be shared if multiple fields share the same type.  Or possibly up to the constructor level for the read-time ones?  I fear pulling up the write-time ones since it may have more info about the types at runtime and it may vary per call.  Not sure how that interacts with erasure yet...

### DIFF
--- a/integration-test-java/src/main/java/com/vimeo/sample_java_model/DynamicallyTypedModel.java
+++ b/integration-test-java/src/main/java/com/vimeo/sample_java_model/DynamicallyTypedModel.java
@@ -1,0 +1,27 @@
+package com.vimeo.sample_java_model;
+
+import com.google.gson.reflect.TypeToken;
+import com.vimeo.stag.UseStag;
+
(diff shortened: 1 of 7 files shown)


mcumings:
Superseded by #161.

anirudhramanan:
@mcumings https://github.com/vimeo/stag-java/pull/161 should fix the wildcard issue. kindly review

anirudhramanan:
didn't see the previous comment, my bad

PR #167, opened 2018-02-27
Using PODAM to run some integration tests on type adapter output

anthonycr:
# Summary
Using PODAM to run some integration tests on type adapter output in the `integration-test-java` module. These integration tests are only for a few members of the package. In order for these tests to work, `equals` must be properly implemented, which resulted in some members needing `equals` to be added. Note that PODAM is already being used in the Kotlin module to verify type adapters that are generated.

More tests will be added in the future, but right now only some of the models are being tested in the `integration-test-java` module due to needing to spend some time helping PODAM generate data for certain models.

## How to Test
Let the tests run.


integration-test-java/src/test/java/verification/Utils.java
```diff
@@ -71,4 +77,30 @@ private Utils() {
...
+     * @param clazz the {@link Class} to use to get the {@link TypeAdapter}.
+     */
+    public static <T> void verifyTypeAdapterCorrectness(@NotNull Class<T> clazz) {
+        final PodamFactory factory = new PodamFactoryImpl();
+        factory.setClassStrategy(new HungarianNotationClassInfoStrategy());
+
+        final T object = factory.manufacturePojo(clazz);
+        final TypeAdapter<T> typeAdapter = getTypeAdapter(clazz);
+
+
```
brentwatson: nit: extra new line.

integration-test-java/src/test/java/verification/Utils.java
```diff
@@ -71,4 +77,30 @@ private Utils() {
...
+        T newObject = null;
+        if (typeAdapter != null) {
+            String json = typeAdapter.toJson(object);
```
brentwatson: can be `final`

PR #175, opened 2018-04-23
Adding more tests around the adapters

anthonycr:
# Summary
Increasing test coverage of the `stag-library` module.

## How to Test
- Run the tests.

### DIFF
--- a/stag-library/src/test/java/com/vimeo/stag/KnownTypeAdaptersTest.java
+++ b/stag-library/src/test/java/com/vimeo/stag/KnownTypeAdaptersTest.java
@@ -392,7 +392,7 @@ public class KnownTypeAdaptersTest {
     }
 
     @Test
-    public void primitiveArrayFloatTypeAdapterHandlesNullsCorrectly() {
+    public void primitiveArrayFloatTypeAdapterHandlesNullsCorrectly() throws Exception {
         final float[] input = null;
         StringWriter stringWriter = new StringWriter();
         KnownTypeAdapters.PrimitiveFloatArrayAdapter.write(new JsonWriter(stringWriter), input);
--- a/stag-library/src/test/java/com/vimeo/stag/KnownTypeAdaptersTest.java
+++ b/stag-library/src/test/java/com/vimeo/stag/KnownTypeAdaptersTest.java
@@ -611,8 +611,9 @@ public class KnownTypeAdaptersTest {
         // for string arrays
         ArrayList<String> dummyList = Utils.createStringDummyList();
 
-        TypeAdapter<ArrayList<String>> listTypeAdapter = new KnownTypeAdapters.ListTypeAdapter<>(TypeAdapters.STRING,
-                                                                                                 new KnownTypeAdapters.ArrayListInstantiator<String>());
+        TypeAdapter<ArrayList<String>> listTypeAdapter =
+                new KnownTypeAdapters.ListTypeAdapter<>(TypeAdapters.STRING,
+                                                        new KnownTypeAdapters.ArrayListInstantiator<String>());
 
         StringWriter stringWriter = new StringWriter();
         listTypeAdapter.write(new JsonWriter(stringWriter), dummyList);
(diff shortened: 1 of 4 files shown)


stag-library/src/test/java/com/vimeo/stag/KnownTypeAdaptersTest.java
```diff
@@ -542,6 +646,22 @@ public void testForListTypeAdapter() throws Exception {
         }
     }
 
+    @Test
+    public void listTypeAdapterHandlesNullsCorrectly() throws Exception {
+        final ArrayList<String> dummyList = null;
+
+        TypeAdapter<ArrayList<String>> listTypeAdapter = new KnownTypeAdapters.ListTypeAdapter<>(TypeAdapters.STRING,
```
brentwatson: nit: long line. (Same with L709)

stag-library/src/test/java/com/vimeo/stag/KnownTypeAdaptersTest.java
```diff
@@ -582,6 +702,22 @@ public void testForMapTypeAdapter() throws Exception {
...
     }
 
+    @Test
+    public void mapTypeAdapterHandlesNullsCorrectly() throws Exception {
+        final HashMap<String, String> dummyMap = null;
+
+        TypeAdapter<HashMap<String, String>> mapTypeAdapter = new KnownTypeAdapters.MapTypeAdapter<>(TypeAdapters.STRING, TypeAdapters.STRING,
+                                                                                                     new KnownTypeAdapters.HashMapInstantiator<String, String>());
+
+        StringWriter stringWriter = new StringWriter();
```
brentwatson: This `StringWriter` setup + adapter execution test code is repeated a lot throughout these tests. You might want to consider making some private local util method. However, since the adapter is the thing under test if you want to keep all of this test-specific, that's OK too. Your call.

stag-library/src/test/java/com/vimeo/stag/KnownTypeAdaptersTestKt.kt
```diff
@@ -37,6 +38,11 @@ class KnownTypeAdaptersTestKt {
         KnownTypeAdapters.PrimitiveByteTypeAdapter.read(JsonReader(StringReader("test".quote())), 1)
     }
 
+    @Test
+    fun `PrimitiveByteTypeAdapter is not instantiable`() {
+        testZeroArgumentConstructorFinalClass(KnownTypeAdapters.PrimitiveByteTypeAdapter::class.java)
```
brentwatson: You could make `T` `reified` in `testZeroArgumentConstructorFinalClass` so that this can can be written a bit nicer (`testZeroArgumentConstructorFinalClass<KnownTypeAdapters.PrimitiveByteTypeAdapter>()`)

eg, something like:
```kotlin
inline fun <reified T> testZeroArgumentConstructorFinalClass() {
    assertThatThrownBy {
        T::class.java.getDeclaredConstructor().apply {
            isAccessible = true
        }.newInstance()
    }.hasCauseInstanceOf(UnsupportedOperationException::class.java)
}
```

stag-library/src/test/java/com/vimeo/stag/Utils.kt
```diff
@@ -5,23 +5,20 @@ package com.vimeo.stag
...
-        val constructor = clazz.getDeclaredConstructor()
-        constructor.isAccessible = true
-        constructor.newInstance()
-    } catch (e: InvocationTargetException) {
-        if (e.cause is UnsupportedOperationException) {
-            exceptionThrown = true
-        }
+    fun instantiateClass(clazz: Class<T>): T {
+        return clazz.getDeclaredConstructor().apply {
+            isAccessible = true
```
brentwatson: Right before this you could also add a `assertFalse(isAccessible)` to ensure the class is defined correctly before you modify it.

stag-library/src/test/java/com/vimeo/stag/KnownTypeAdaptersTest.java
anthonycr: I noticed this as well, but I am going to hold off doing the refactoring to abstract this until I convert this class to Kotlin, because I tried it out and it creates more code in java and makes the tests more difficult to understand. Here's what I got:
```java
private <T> T writeAndReadValue(T value, Function2<JsonWriter, T, Unit> write, Function1<JsonReader, T> read) {
    StringWriter stringWriter = new StringWriter();
    write.invoke(new JsonWriter(stringWriter), value);
    String jsonString = stringWriter.toString();

    return read.invoke(new JsonReader(new StringReader(jsonString)));
}

// usage
@Test
public void testForPrimitiveByteTypeAdapter() throws Exception {
    byte value = 1;

    byte readValue = writeAndReadValue(value, new Function2<JsonWriter, Byte, Unit>() {
        @Override
        public Unit invoke(JsonWriter jsonWriter, Byte aByte) {
            try {
                KnownTypeAdapters.PrimitiveByteTypeAdapter.write(jsonWriter, aByte);
            } catch (IOException e) {
                throw new RuntimeException(e);
            }
            return Unit.INSTANCE;
        }
    }, new Function1<JsonReader, Byte>() {
        @Override
        public Byte invoke(JsonReader jsonReader) {
            try {
                return KnownTypeAdapters.PrimitiveByteTypeAdapter.read(jsonReader, (byte) 0);
            } catch (IOException e) {
                throw new RuntimeException(e);
            }
        }
    });

    assertEquals(value, readValue);
}
```

anthonycr:
@brentwatson thanks for review, I made the changes. Back to you for re-review.

PR #177, opened 2018-06-15
Unit tests

anthonycr:
# Summary
Adding more unit tests which mainly cover utils. Also cleaned up some code and removed unused functions.

## How to test
Let the tests run.

### DIFF
--- a/stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/TypeUtils.java
+++ b/stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/TypeUtils.java
@@ -439,8 +439,9 @@ public final class TypeUtils {
     /**
      * Method to check if the {@link TypeMirror} is of primitive type
      *
-     * @param type :TypeMirror type
-     * @return boolean
+     * @param type TypeMirror type
+     * @return true if the type represented as a string is a primitive type that's supported, false
+     * otherwise.
      */
     public static boolean isSupportedPrimitive(@NotNull String type) {
         return PRIMITIVE_TO_OBJECT_MAP.containsKey(type);
@@ -449,8 +450,8 @@ public final class TypeUtils {
     /**
      * Method to check if the {@link TypeMirror} is of {@link ArrayType}
      *
-     * @param type :TypeMirror type
-     * @return boolean
+     * @param type TypeMirror type
+     * @return true if the type is a native array type, false otherwise.
      */
     public static boolean isNativeArray(@NotNull TypeMirror type) {
         return (type instanceof ArrayType);
@@ -460,7 +461,8 @@ public final class TypeUtils {
      * Method to check if the {@link TypeMirror} is of {@link List} type
      *
      * @param type TypeMirror type
-     * @return boolean
+     * @return true if the type is a {@link List}, {@link ArrayList}, or {@link Collection},
+     * false otherwise.
      */
     public static boolean isSupportedList(@NotNull TypeMirror type) {
         String outerClassType = TypeUtils.getOuterClassType(type);
@@ -472,8 +474,8 @@ public final class TypeUtils {
     /**
      * Method to check if the {@link TypeMirror} is of {@link Map} type
      *
-     * @param type :TypeMirror type
-     * @return boolean
+     * @param type TypeMirror type
+     * @return true if the type is a supported map type, false otherwise.
      */
     public static boolean isSupportedMap(@Nullable TypeMirror type) {
         if (type == null) {
@@ -555,7 +557,7 @@ public final class TypeUtils {
     /**
      * Return the type of JsonAdapter {@link TypeMirror}
      *
-     * @param type :TypeMirror type
+     * @param type TypeMirror type
      * @return {@link JsonAdapterType}
      */
     @NotNull
--- a/stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/TypeUtils.java
+++ b/stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/TypeUtils.java
@@ -28,6 +28,7 @@ import com.google.gson.JsonSerializer;
 import com.google.gson.TypeAdapter;
 import com.google.gson.TypeAdapterFactory;
 import com.vimeo.stag.processor.generators.model.accessor.FieldAccessor;
+import com.vimeo.stag.processor.utils.logging.DebugLog;
 
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
--- a/stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/TypeUtils.java
+++ b/stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/TypeUtils.java
@@ -273,8 +273,8 @@ public final class TypeUtils {
      * inherits from Object or Enum.
      */
     @Nullable
-    public static TypeMirror getInheritedType(@Nullable TypeElement element) {
-        TypeMirror typeMirror = element != null ? element.getSuperclass() : null;
+    public static TypeMirror getInheritedType(@NotNull TypeElement element) {
+        TypeMirror typeMirror = element.getSuperclass();
         String className = typeMirror != null ? getClassNameFromTypeMirror(typeMirror) : null;
         if (!Object.class.getName().equals(className) && !Enum.class.getName().equals(className)) {
             return typeMirror;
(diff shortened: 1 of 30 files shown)


stag-library-compiler/src/main/java/com/vimeo/stag/processor/utils/TypeUtils.java
```diff
@@ -456,26 +456,13 @@ public static boolean isNativeArray(@NotNull TypeMirror type) {
...
-    public static boolean isSupportedCollection(@Nullable TypeMirror type) {
-        return type != null && (isNativeArray(type) || isSupportedList(type));
-    }
-
     /**
      * Method to check if the {@link TypeMirror} is of {@link List} type
      *
-     * @param type :TypeMirror type
+     * @param type TypeMirror type
      * @return boolean
```
brentwatson: optional: you could change the 4 `@return boolean` docs in this class to be more descriptive instead of just declaring the return type.  Eg: `@return true if the given type is a List, otherwise false`.

stag-library-compiler/src/test/java/com/vimeo/stag/processor/ElementUtilsUnitTest.kt
```diff
@@ -48,6 +47,16 @@ class ElementUtilsUnitTest : BaseUnitTest() {
 
     @Test
     fun testGetTypeFromQualifiedName() {
+        fun <T> testTypeMirrorCreationAndEquality(clazz: Class<T>) {
```
brentwatson: 1. Only tests that doesn't use a "nice" name (method name with spaces).
2. Might actually be worth breaking this up into 2 tests (`fun test creation equality` and `fun test creation equality`), since testing equality and inequality are 2 different things.
3. Optional; might be easier to mentally parse something like:
```kotlin
listOf(
    String::class.java,
    Any::class.java,
    ArrayList::class.java,
    DummyConcreteClass::class.java,
    DummyGenericClass::class.java,
    DummyInheritedClass::class.java
).forEach { clazz ->
    assertEquals(Utils.getTypeMirrorFromClass(clazz), ElementUtils.getTypeFromQualifiedName(clazz.name))
}
```
(not sure what the nested function offers over this).

stag-library-compiler/src/test/java/com/vimeo/stag/processor/TypeUtilsUnitTest.kt
```diff
@@ -285,113 +309,40 @@ class TypeUtilsUnitTest : BaseUnitTest() {
...
+        assertTrue(TypeUtils.isSupportedMap(Utils.getTypeMirrorFromClass(LinkedHashMap::class.java)))
 
         // Check type that implements map
-        Assert.assertFalse(TypeUtils.isSupportedMap(Utils.getTypeMirrorFromClass(DummyMapClass::class.java)))
+        assertFalse(TypeUtils.isSupportedMap(Utils.getTypeMirrorFromClass(DummyMapClass::class.java)))
 
         // Check other types
-        Assert.assertFalse(TypeUtils.isSupportedMap(Utils.getTypeMirrorFromClass(Any::class.java)))
-        Assert.assertFalse(TypeUtils.isSupportedMap(Utils.getTypeMirrorFromClass(DummyConcreteClass::class.java)))
-        Assert.assertFalse(TypeUtils.isSupportedMap(Utils.getTypeMirrorFromClass(String::class.java)))
```
brentwatson: I like this removal of `Assert.`. There are also 4 remaining `Assert.assertXXXX` in SwappableParserExampleModelTest.java that could also be changed if you want to convert them all while you're at it.

stag-library-compiler/src/test/java/com/vimeo/stag/processor/utils/DebugLogTest.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    @Test
+    fun `DebugLog is not instantiable`() {
+        Utils.testZeroArgumentConstructorFinalClass(DebugLog::class.java)
+    }
+
+    @Test
+    fun `log message works`() {
+        StagProcessor.DEBUG = true
+        DebugLog.log("test message")
```
brentwatson: This isn't really testing that logging is working. It's only testing that exceptions aren't being thrown when trying to log.

If it's worth actually testing this, then you could make `DebugLog` hold a `OutputStream` variable which defaults to `System.out` and mock an `OutputStream` during testing.

If you do this then also add some tests that test when `StagProcessor.DEBUG = false`.

stag-library-compiler/src/test/java/com/vimeo/stag/processor/utils/FileGenUtilsTest.kt
```diff
@@ -0,0 +1,40 @@
...
+
+    @Test
+    fun `FileGenUtils is not instantiable`() {
+        Utils.testZeroArgumentConstructorFinalClass(FileGenUtils::class.java)
+    }
+
+    @Test
+    fun close() {
+        var counter = 0
+        val closeable = Closeable { counter++ }
```
brentwatson: 🏆 nice.

stag-library-compiler/src/test/java/com/vimeo/stag/processor/utils/FileGenUtilsTest.kt
```diff
@@ -0,0 +1,40 @@
...
+        FileGenUtils.close(closeable)
+        assertEquals(1, counter)
+
+        FileGenUtils.close(null)
+        assertEquals(1, counter)
+
+        val failureCloseable = Closeable { throw IOException("test") }
```
brentwatson: This could be in it's own test.

stag-library/src/main/java/com/vimeo/stag/Types.java
```diff
@@ -1,35 +1,41 @@
 package com.vimeo.stag;
```
brentwatson: nit: extra blank line can be removed.

stag-library/src/main/java/com/vimeo/stag/Types.java
```diff
@@ -1,35 +1,41 @@
...
 import java.lang.reflect.WildcardType;
 
+@SuppressWarnings("unused")
 public final class Types {
 
     private Types() {}
 
-    public static WildcardType getWildcardType(Type[] upperBounds, Type[] lowerBounds) {
+    @NotNull
+    public static WildcardType getWildcardType(@NotNull Type[] upperBounds, @NotNull Type[] lowerBounds) {
```
brentwatson: 🏆 nice to see all the nullability annotations added in this PR!

stag-library/src/test/java/com/vimeo/stag/KnownTypeAdaptersTestKt.kt
```diff
@@ -268,4 +300,81 @@ class KnownTypeAdaptersTestKt {
...
+                ObjectConstructor<Map<KeyTest, String>> { HashMap() }
+        )
+
+        val testMap = mapOf(
+                KeyTest("a") to "b",
+                KeyTest("b") to "c"
+        )
+
+        val json = mapTypeAdapter.toJson(testMap)
+        assertThat(mapTypeAdapter.fromJson(json)).isEqualTo(testMap)
```
brentwatson: I didn't dig into this too much, but on the surface it looks like you are only testing the `mapTypeAdapter` that you create in the test (with overridden `read` and `write` methods). Is there any part of `MapTypeAdapter` that is actually being tested here? If so might be worth adding an inline comment somewhere explaining what part of `MapTypeAdapter` is being tested. Same applies to the next unit test.

stag-library-compiler/src/test/java/com/vimeo/stag/processor/utils/DebugLogTest.kt
anthonycr: Made `DebugLog` take a logging implementation and then injecting that when the processor gets set up. It is now testable, good callout.

stag-library/src/test/java/com/vimeo/stag/KnownTypeAdaptersTestKt.kt
anthonycr: It's a little difficult to tell because the instantiation is verbose, but the `read` and `write` methods are from the `TypeAdapter` that is being _injected_ into the `MapTypeAdapter`. So these tests are verifying that the adapter properly delegates to the injected `TypeAdapter` depending on the type used. I broke out the adapter into its own variable to make it more clear.

Issue #181, opened 2018-08-01
class generated between compiles is different


pouloghost:
#### Issue Summary
generated classes depend on         
`Set<? extends Element> useStagElements = roundEnv.getElementsAnnotatedWith(UseStag.class);`
can vary between compiles, making hotfix very hard.

#### Reproduction Steps
just compile several times

#### Expected Behavior
generated classes stay the same unless code changes.

#### Fix step
`
       Set<? extends Element> useStagElements = roundEnv.getElementsAnnotatedWith(UseStag.class);
        List<? extends Element> sorted = new ArrayList<>(useStagElements);
        sorted.sort(new Comparator<Element>() {
            @Override
            public int compare(Element element, Element t1) {
                return element.toString().compareTo(t1.toString());
            }
        });
`

anthonycr:
Great find, I dug into the code and discovered it is due to the usage of `HashSet` and `HashMap` in the code. Neither of these collections guarantee iteration order, and therefore cause builds to be non-deterministic. The fix was to switch to `LinkedHashSet` and `LinkedHashMap` for all usages of `Set` and `Map` that were iterated.

I did not perform sorting of the elements returned by `getElementsAnnotatedWith` as it is deterministic and returns the elements in a fixed order for the same input. I verified this additionally by adding print statements of `useStagElements.toString().hashCode()`, and observed that it printed the same value for every invocation of `./gradlew clean :sample:assembleDebug`.

Bug fixed here: https://github.com/vimeo/stag-java/pull/184

anthonycr:
Closing as fixed: https://github.com/vimeo/stag-java/pull/184

PR #184, opened 2018-09-10
Deterministic compilation

anthonycr:
# Issue 
- https://github.com/vimeo/stag-java/issues/181

## Summary
Compilation was not deterministic due to usage of `HashSet` and `HashMap`. Switched to `LinkedHashMap` and `LinkedHashSet` in order to guarantee deterministic iteration and deterministic compilation. Added a functional test that validates consistent output between compilations. 

## How to Test
Let the tests run. See that the old test fails if the changes to the usage of the implementation of `Map` and `Set` are reverted.


stag-library-compiler/src/test/java/com/vimeo/stag/processor/functional/JavaIntegrationFunctionalTests.kt
```diff
@@ -150,8 +151,49 @@ class JavaIntegrationFunctionalTests {
...
+
+        assertThat(compilation1Hash).isEqualTo(compilation2Hash)
+    }
+
+    /**
+     * Returns the concatenation of the hash of each file generated by the [Compilation].
+     */
+    private fun Compilation.hashOutput(): String = generatedFiles()
+            .map { it.getCharContent(false).hashCode() }
+            .joinToString()
```
brentwatson: Optional: `joinToString` can take a `transform`, so you could move the logic from `map` and just use the `joinToString`... 

```kotlin
.joinToString { it.getCharContent(false).hashCode().toString() }`
```

